PR #2, opened 2020-07-21
Support for certs and mounting them.

bigkevmcd:
Add support for creating SSL certs for services and mounting them into the container.

### DIFF
--- a/pkg/controller/gitopsservice/gitopsservice_controller.go
+++ b/pkg/controller/gitopsservice/gitopsservice_controller.go
@@ -201,16 +201,18 @@ func (r *ReconcileGitopsService) Reconcile(request reconcile.Request) (reconcile
 	return reconcile.Result{}, nil
 }
 
-func objectMeta(resourceName string, namespace string) metav1.ObjectMeta {
-	ObjectMeta := metav1.ObjectMeta{
+func objectMeta(resourceName string, namespace string, opts ...func(*metav1.ObjectMeta)) metav1.ObjectMeta {
+	objectMeta := metav1.ObjectMeta{
 		Name:      resourceName,
 		Namespace: namespace,
 	}
-	return ObjectMeta
+	for _, o := range opts {
+		o(&objectMeta)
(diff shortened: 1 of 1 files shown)


pkg/controller/gitopsservice/gitopsservice_controller.go
```diff
@@ -276,14 +295,17 @@ func newServiceForCR(cr *pipelinesv1alpha1.GitopsService) *corev1.Service {
 		},
 	}
 	svc := &corev1.Service{
-		ObjectMeta: objectMeta(cr.Name, cr.Namespace),
-		Spec:       spec,
+		ObjectMeta: objectMeta(cr.Name, cr.Namespace, func(o *metav1.ObjectMeta) {
+			o.Annotations = map[string]string{
+				"service.beta.openshift.io/serving-cert-secret-name": cr.Name,
```
sbose78, line 300: Could you point me to docs for this, please?

bigkevmcd: https://docs.openshift.com/container-platform/4.5/security/certificates/service-serving-certificate.html#add-service-certificate_service-serving-certificate

pkg/controller/gitopsservice/gitopsservice_controller.go
```diff
@@ -229,6 +231,23 @@ func newDeploymentForCR(cr *pipelinesv1alpha1.GitopsService) *appsv1.Deployment
...
+					},
+				},
+			},
+		},
+		Volumes: []corev1.Volume{
+			{
+				Name: "backend-ssl",
+				VolumeSource: corev1.VolumeSource{
+					Secret: &corev1.SecretVolumeSource{
+						SecretName: "pipelines-app-delivery-backend",
```
chetan-rns, line 248: The secret name should be `cr.Name`.  It should match the value provided in the annotation `"service.beta.openshift.io/serving-cert-secret-name"`

PR #8, opened 2020-09-09
Change the CRD version from v1 to v1beta1

chetan-rns:
Currently, the community operators don't support CRD version `apiextensions.k8s.io/v1`. We need to change the version back to `apiextensions.k8s.io/v1beta1` which was the case with earlier versions of `operator-sdk` (<18.0). I updated the CRDs with the following approach:

1. Install `operator-sdk` v0.17.0 locally.
2. Use `operator-sdk generate crds` to update the CRDs
3. Verify if the operator is working as expected

PR to add GItOps operator to OperatorHub: https://github.com/operator-framework/community-operators/pull/2293

Note: I did try to downgrade the `operator-sdk` version. Later inspected the ArgoCD operator and figured out it uses the latest version of `operator-sdk` and the only modified files are CRDs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ That's it, your API `route` should be created for you. You don't need to explicl
 2. Login to a cluster on your command-line.
 3. `OPERATOR_NAME=gitops-operator operator-sdk run local --watch-namespace=openshift-pipelines-app-delivery`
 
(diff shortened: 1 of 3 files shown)


bigkevmcd (approved):
Thanks for this, I wonder if we should add something to the README.md warning about this?

bigkevmcd (approved):
Thanks !

sbose78 (approved):

PR #9, opened 2020-10-01
Create console link from the operator

chetan-rns:
ConsoleLink is a cluster-scoped CRD used by admins to customize OpenShift console. Installing the GitOps operator should create a ConsoleLink resource, which will be useful to navigate to external ArgoCD UI. This PR adds a new controller that can reconcile on ArgoCD instance in `argocd ` namespace and handle ConsoleLink resource.

**Watchers:**

```
1. ArgoCD instance:  Reconcile only for ArgoCD instance named argocd in argocd namespace. We can ignore other instances of ArgoCD
2. ArgoCD server route: Reconcile when the ArgoCD server route is created/updated/deleted and regenerate the ConsoleLink 
```

**Controller Logic:**
```

1. Fetch ArgoCD instance
2. If the instance is absent
        * delete the ConsoleLink if present
3. Fetch the argocd-server route
4. If the route is absent
        * delete the ConsoleLink if present
5. Create the ConsoleLink if absent 

```

**Steps to test:**

1. Install the ArgoCD operator in `argocd` namespace and create an instance named `argocd` with route enabled
2. Install the GitOps operator 
3. Check the ConsoleLink in the `Applications` menu

Screenshot from 2020-10-05 17-14-59


chetan-rns:
> You could also check if the `resourceVersion` of the object has changed, the event should be ignored.

Yeah. I've added a check to ignore if the resource version is not changed

go.mod
```diff
@@ -3,12 +3,14 @@ module github.com/redhat-developer/gitops-operator
 go 1.13
 
 require (
-	github.com/openshift/api v3.9.0+incompatible
+	github.com/argoproj-labs/argocd-operator v0.0.13
+	github.com/go-logr/logr v0.1.0
+	github.com/google/go-cmp v0.4.0
+	github.com/openshift/api v3.9.1-0.20190916204813-cdbe64fb0c91+incompatible
 	github.com/operator-framework/operator-sdk v0.18.2
-	github.com/prometheus/common v0.9.1
```
wtam2018, line 8: Can we remote the following under `replace` now that `prometheus` is removed?

```
k8s.io/client-go => k8s.io/client-go v0.18.2 // Required by prometheus-operator
```

wtam2018 (approved):
LGTM
Please follow up my comment in go.mod to see if it is safe to remove the "replace" statement.

go.mod
chetan-rns: @wtam2018 Tried removing this replace. It is throwing errors. It may be dependent on the operator-sdk version

chetan-rns:
@bigkevmcd @wtam2018 Can you PTAL? Please let me know if I need to change anything

PR #10, opened 2020-10-23
Install dependencies from GitOps operator

chetan-rns:
The GitOps operator should be capable of configuring and installing GitOps specific dependencies on your OpenShift cluster. This will eliminate the hassle of manual configuration for the argocd/pipelines operator.

Since OLM doesn't support cross-namespace dependency resolution, only the pipeline operator will be installed by OLM. The GitOps operator will manually create namespace (argocd) and install ArgoCD operator by creating operator groups and subscriptions. 

**Workflow:**

1. User will install the GitOps operator from OperatorHub. OLM will install the dependent OpenShift pipeline operator.
2. The reconciler will  install ArgoCD  operator in the `argocd` namespace.
4. The reconciler will then create the CR for the ArgoCD  operator.

Co-authored by: @gaganhegde 
 
**Important Note :** Since the sealed secret operator cannot handle multiple instances, this PR will only install the  ArgoCD and Pipelines Operator as dependencies.


wtam2018 (dismissed):
LGTM
Please follow up my comment in go.mod to see if it is safe to remove the "replace" statement.

deploy/olm-catalog/gitops-operator/manifests/gitops-operator.clusterserviceversion.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
 metadata:
   annotations:
-    containerImage: quay.io/redhat-developer/gitops-backend-operator:v0.0.1
+    containerImage: quay.io/cbanavik/gitops-backend-operator:dep
```
bigkevmcd: I don't think this should be in this PR?

pkg/apis/pipelines/v1alpha1/gitopsservice_types.go
```diff
@@ -12,6 +12,7 @@ type GitopsServiceSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
```
bigkevmcd, line 13: Remove this comment, and add a proper comment for the Prefix field, it's embedded in the CRD.

https://github.com/redhat-developer/gitops-operator/pull/10/commits/a27b74582dd4544e860b31eb936a9c197e3c2018#diff-4284bda429d34d9eb75be67605112b07f66752910ea60847e9980bec6b9afccfR35

deploy/operator.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       containers:
         - name: gitops-operator
           # Replace this with the built image name
-          image: REPLACE_IMAGE
+          image: quay.io/cbanavik/gitops-operator:role
```
bigkevmcd: I think this is automatically replaced, and so we likely don't want your test image.

pkg/controller/gitopsservice/dependency/dependency.go
```diff
@@ -0,0 +1,143 @@
...
+		// TODO: Create the operator CR
+	}
+
+	return nil
+}
+
+func (d *Dependency) createResourceIfAbsent(ctx context.Context, obj runtime.Object, ns types.NamespacedName) error {
+	err := d.client.Get(ctx, ns, obj)
+	if err != nil {
+		if errors.IsNotFound(err) {
```
bigkevmcd, line 85: You could probably improve this by using an expression switch and cases for the options (and `default:` returning the error).

bigkevmcd (changes requested):
Reading the image will fail, but that's being addressed, and there are a number of TODOs that it's not clear _when_ they'll be addressed?

deploy/olm-catalog/gitops-operator/manifests/gitops-operator.clusterserviceversion.yaml
chetan-rns: This is an outdated commit. It was changed earlier

deploy/operator.yaml
chetan-rns: This is outdated as well. I shouldn't have pushed it in the first place :)

deploy/crds/pipelines.openshift.io_v1alpha1_gitopsservice_cr.yaml
```diff
@@ -5,3 +5,4 @@ metadata:
 spec:
   # Add fields here
   size: 3
```
bigkevmcd, line 7: I think we can remove the `size` field, it's not used by the reconciler.

cmd/manager/main.go
```diff
@@ -110,7 +112,7 @@ func main() {
 
 	// Set default manager options
 	options := manager.Options{
-		Namespace:          namespace,
+		Namespace:          "",
 		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
 	}
```
wtam2018, line 118: Can we get rid of the multi namespace support below?

pkg/controller/gitopsservice/dependency/dependency_test.go
```diff
@@ -0,0 +1,179 @@
...
+	"context"
+	"fmt"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	argo "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
+	v1 "github.com/operator-framework/api/pkg/operators/v1"
+	"github.com/operator-framework/api/pkg/operators/v1alpha1"
```
wtam2018, line 13: run `go mod tidy`

cmd/manager/main.go
chetan-rns: The GitOps operator only supports the `all-namespace` installation mode and it creates resources across multiple namespaces. If the CR is namespace scoped, the controller would only create resources in the namespace of the CR

wtam2018: I mean can this block of code safely removed?
```
	// Add support for MultiNamespace set in WATCH_NAMESPACE (e.g ns1,ns2)
	// Note that this is not intended to be used for excluding namespaces, this is better done via a Predicate
	// Also note that you may face performance issues when using this with a high number of namespaces.
	// More Info: https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/cache#MultiNamespacedCacheBuilder
	if strings.Contains(namespace, ",") {
		options.Namespace = ""
		options.NewCache = cache.MultiNamespacedCacheBuilder(strings.Split(namespace, ","))
	}
```

pkg/controller/gitopsservice/config/config.go
```diff
@@ -0,0 +1,64 @@
...
+	return Config{c}
+}
+
+// ExtractPrefixes will extract the prefixes list from config
+func (c *Config) ExtractPrefixes() []string {
+	prefixes := []string{}
+	if prefixString, ok := c.Data[gitopsPrefix]; ok {
+		prefixes = append(prefixes, strings.Split(prefixString, ",")...)
+	}
+	return prefixes
```
wtam2018, line 45:
```suggestion
	prefixes := []string{}
	if prefixString, ok := c.Data[gitopsPrefix]; ok {
		return strings.Split(prefixString, ",")
	}
	return []string{}
```

pkg/controller/gitopsservice/dependency/dependency.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+func addPrefixIfNecessary(prefix, name string) string {
+	if prefix != "" {
+		return prefix + "-" + name
+	}
+	return name
+}
+
+func argoCDCr(ns string) (runtime.Object, string, error) {
```
wtam2018, line 209: function nane should be changed to `argoCDCR`.

pkg/controller/gitopsservice/dependency/dependency.go
```diff
@@ -0,0 +1,269 @@
...
+		Object: map[string]interface{}{
+			"apiVersion": "bitnami.com/v1alpha1",
+			"kind":       "SealedSecretController",
+			"metadata": map[string]interface{}{
+				"name":      name,
+				"namespace": ns,
+			},
+			"spec": map[string]interface{}{
+				"image": map[string]interface{}{
+					"repository": "quay.io/bitnami/sealed-secrets-controller@sha256:8e9a37bb2e1a6f3a8bee949e3af0e9dab0d7dca618f1a63048dc541b5d554985",
```
wtam2018, line 251: We need to be able get image from environment variable in order to support disconnected install.

cmd/manager/main.go
chetan-rns: Removed the unwanted code

pkg/controller/gitopsservice/dependency/dependency.go
chetan-rns: Added the support to read images from env variable

chetan-rns:
@bigkevmcd @wtam2018 Removed sealed secret operator dependency. Please consider it for review

bigkevmcd (approved):

PR #14, opened 2020-11-17
Update the controller to reconcile for route instead of ArgoCD instance

chetan-rns:
In order to reconcile for the ArgoCD CR instance, the controller expects the ArgoCD CRD to be present beforehand. In the absence of CRD, the controller wouldn't be created. This PR addresses this issue by changing the controller logic to reconcile for the ArgoCD server route. Since Route is a native OpenShift object, we can expect the CRD to be already present.


pkg/controller/argocd/argocd_controller.go
```diff
@@ -137,39 +110,20 @@ type ReconcileArgoCD struct {
...
-			return reconcile.Result{}, r.deleteConsoleLinkIfPresent(ctx, reqLogger)
-		}
-		// Error reading the object - requeue the request.
-		return reconcile.Result{}, err
-	}
-
-	reqLogger.Info("ArgoCD instance found", "ArgoCD.Namespace:", argocdInstance.Namespace, "ArgoCD.Name", argocdInstance.Name)
-
-	// Set ArgoCD instance as the owner
-	if err := controllerutil.SetControllerReference(argocdInstance, newArgoCDRoute(), r.scheme); err != nil {
```
bigkevmcd, line 167: You can remove `newArgoCDRoute` now?

pkg/controller/argocd/argocd_controller.go
```diff
@@ -66,36 +63,16 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 	reqLogger := logs.WithValues()
-	reqLogger.Info("Watching ArgoCD")
+	reqLogger.Info("Watching ArgoCD Server Route")
 
-	// Skip controller creation if ArgoCD CRD is not present
-	_, err := mgr.GetRESTMapper().RESTMapping(schema.GroupKind{
-		Group: argocdGroup,
```
bigkevmcd, line 73: You don't need either of these constants now?

`argocdKind` or `argocdGroup`

pkg/controller/argocd/argocd_controller.go
```diff
@@ -118,10 +95,6 @@ func filterPredicate(assert func(namespace, name string) bool) predicate.Funcs {
 	}
 }
 
-func assertArgoCD(namespace, name string) bool {
-	return namespace == argocdNS && argocdInstanceName == name
-}
-
 func assertArgoCDRoute(namespace, name string) bool {
```
bigkevmcd, line 98: nit: I'd have called this `argoCDRouteFilter`

pkg/controller/argocd/argocd_controller.go
```diff
@@ -137,39 +110,20 @@ type ReconcileArgoCD struct {
...
 }
 
-// Reconcile reads that state of the cluster for a ArgoCD object and makes changes based on the state read
-// and what is in the ArgoCD.Spec
+// Reconcile reads that state of the cluster for a ArgoCD Route object and makes changes based on the state read
+// and what is in the Route.Spec
 // Note:
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileArgoCD) Reconcile(request reconcile.Request) (reconcile.Result, error) {
```
bigkevmcd, line 118: Should this be renamed, now that it's no longer reconciling ArgoCD objects?

pkg/controller/argocd/argocd_controller.go
```diff
@@ -178,26 +132,29 @@ func (r *ReconcileArgoCD) Reconcile(request reconcile.Request) (reconcile.Result
 		}
 		return reconcile.Result{}, err
 	}
-
 	reqLogger.Info("Route found for argocd-server", "Route.Host", argoCDRoute.Spec.Host)
 
-	consoleLink := newConsoleLink("https://"+argoCDRoute.Spec.Host, "ArgoCD")
+	argocCDRouteURL := fmt.Sprintf("https://%s", argoCDRoute.Spec.Host)
```
bigkevmcd, line 137: There's probably nothing we can do here, but it feels weird for us to be calculating the route URL like this.

bigkevmcd (changes requested):

bigkevmcd:
Looks mostly good, a couple of minor cleanups and let's get this in :-)

pkg/controller/argocd/argocd_controller.go
chetan-rns: Updated

pkg/controller/argocd/argocd_controller.go
chetan-rns: Removed the unused constants

pkg/controller/argocd/argocd_controller.go
chetan-rns: The function name is updated

pkg/controller/argocd/argocd_controller.go
chetan-rns: I've renamed it to `ReconcileArgoCDRoute`. I'm okay with `ReconcileRoute` as well

pkg/controller/argocd/argocd_controller.go
chetan-rns: Yes. I wonder why the protocol is not added in the route object. They externally append while displaying the link on the UI

bigkevmcd (approved):
Looks good.

pkg/controller/argocd/argocd_controller.go
```diff
@@ -118,58 +93,35 @@ func filterPredicate(assert func(namespace, name string) bool) predicate.Funcs {
 	}
 }
 
-func assertArgoCD(namespace, name string) bool {
-	return namespace == argocdNS && argocdInstanceName == name
-}
-
-func assertArgoCDRoute(namespace, name string) bool {
+func argocdRouteFilter(namespace, name string) bool {
```
wtam2018, line 96: `argoCDRouteFilter` for funciton name.

chetan-rns: I've updated the function name to `filterArgoCDRoute`

PR #15, opened 2020-11-18
Read images from env variable to support disconnected cluster

chetan-rns:
Users in disconnected clusters rely on mirror registry and manage images using env variables. Read the image for backend service from an env variable to support disconnected environments.

### DIFF
--- a/pkg/controller/gitopsservice/gitopsservice_controller_test.go
+++ b/pkg/controller/gitopsservice/gitopsservice_controller_test.go
@@ -1,0 +1,40 @@
+package gitopsservice
+
+import (
+	"os"
+	"testing"
+
+	pipelinesv1alpha1 "github.com/redhat-developer/gitops-operator/pkg/apis/pipelines/v1alpha1"
+)
+
+func TestImageFromEnvVariable(t *testing.T) {
+	cr := &pipelinesv1alpha1.GitopsService{}
+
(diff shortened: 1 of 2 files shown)


pkg/controller/gitopsservice/gitopsservice_controller_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	pipelinesv1alpha1 "github.com/redhat-developer/gitops-operator/pkg/apis/pipelines/v1alpha1"
+)
+
+func TestImageFromEnvVariable(t *testing.T) {
+	cr := &pipelinesv1alpha1.GitopsService{}
+
+	t.Run("Image present as env variable", func(t *testing.T) {
+		image := "quay.io/org/test"
+		os.Setenv(backendImageEnvName, image)
```
bigkevmcd, line 15: You should be cleaning this up here (resetting it back to the original value).

bigkevmcd (commented):
Looks good, minor fix for that test.

pkg/controller/gitopsservice/gitopsservice_controller_test.go
chetan-rns: Updated. Using defer to unset env variable

bigkevmcd (approved):

PR #24, opened 2020-12-10
Import argocd operator controller into gitops-operator

sbose78:
## About

Given that this operator is Red Hat-specific, I propose we ship *one* OLM operator containing multiple controllers
* Controller(s) for GitOps Service
* Controller(s) for ArgoCD Operator, added as a Go module dependency.

## Relevant change:
https://github.com/redhat-developer/gitops-operator/pull/24/files#diff-e05913ae216cd147cbbb25685ca0f4d8037c11d403efb0be87dde703705af2af

## Advantages:
*  Users/Customers don't care about the "Gitops Service" individually, they care about the whole package which consists of ArgoCD and the service that powers the UI. 
* This will reduce the number of images - we wouldn't need to build the argocd-operator,  argocd-operator-bundle image anymore.
* This will enable us to keep https://github.com/argoproj-labs/argocd-operator/ as upstream as possible. OpenShift-specific controller changes can live in this repo ( needs more design, but possible )

## Demo

https://www.youtube.com/watch?v=v8-7396xlAM&feature=youtu.be


bigkevmcd (commented):
The changes here are ok, my concern is around the relationship between this code and the upstream argocd-operator.

Given that we've changed to requiring the GitOps operator for the installation of ArgoCD, this feels like a wider discussion is needed.

sbose78:
> Given that we've changed to requiring the GitOps operator for the installation of ArgoCD, this feels like a wider discussion is needed.

Sure, this operator would be the equivalent of "OpenShift Pipelines Operator". That's the only operator on OpenShift Operatorhub that provides Tekton today.

sbose78:
Note, _someone_ may continue releasing argocd operator into [Operatorhub OpenShift Community](framework/community-operators/tree/master/community-operators/argocd-operator) but I would throw more weight around maintaining [Operatorhub Kubernetes Community Operator](https://github.com/operator-framework/community-operators/tree/master/upstream-community-operators/argocd-operator).

> Given that we've changed to requiring the GitOps operator for the installation of ArgoCD

Therefore technically, this PR doesn't necessarily change anything. 


Three operators _could_ be maintained - 

1. OpenShift Gitops Operator 
2.  [Operatorhub OpenShift Community](framework/community-operators/tree/master/community-operators/argocd-operator) 
3. [Operatorhub Kubernetes Community Operator](https://github.com/operator-framework/community-operators/tree/master/upstream-community-operators/argocd-operator)


(2) just feels redundant since we would have (1) anyway.

wtam2018:
@sbose78  I was able to create an ArgoCD CR.   When I created another ArgoCD CR in another namespace, only dex and repo server pods were started.   Argocd controller and api server were not started.  
```
[wtam@localhost gitops-operator]$ k get po
NAME                                   READY   STATUS    RESTARTS   AGE
example-redis-6d88489b48-gjn7v         1/1     Running   0          13m
example-repo-server-7fc8d9758c-9xrrg   1/1     Running   0          13m
```

sbose78:
Things will only work if the controller is in the same namespace - since the changes needed to get to have the controller be useful in all namespaces.

sbose78:
Changes in this PR can be tried out on OpenShift by creating a `CatalogSource` with `quay.io/shbose/gitops-backend-operator-index:v0.0.3`.

All other instructions remain the same https://github.com/redhat-developer/gitops-operator#getting-started

chetan-rns:
Hi @sbose78, +1 for a single GitOps operator. What would be the release process for the GitOps operator? Do we follow the upstream release timeline and keep in sync?

sbose78:
Pulled in changes from argoproj-labs/argocd-operator@8aa649d . Turns out, we will need this commit reverted in argocd-operator :) We'll need to do as per chetan's initial suggestion which is to maintain these clusterrolebindings in the argocd-operator's controller itself.

This is because the name of the clusterrole depends on the OLM operator name “openshift-gitops-operator.v0.0.3-5d467865cf”.

@chetan-rns could you bring back the changes you had removed in https://github.com/argoproj-labs/argocd-operator/pull/221/commits/1403002e2d85efe08c1028e88c74c7d27b674105 ?

chetan-rns:
@sbose78 Couldn't find a better way to extract the cluster roles created by OLM. Will raise a PR to create them from the controller.

sbose78:
Sounds good!

sbose78:
Tested, looks good. This could be merged.

chetan-rns:
@sbose78 The CI is failing because go mod state is not clean. Running go mod tidy and pushing the latest go.mod file may fix this
Screenshot from 2021-01-07 11-38-58

sbose78:
Thank you, pushed again!

chetan-rns (approved):
Thank you!

PR #26, opened 2021-01-06
Fix failing unit tests

chetan-rns:
- Remove cluster roles from CSV file as they're being managed by the controller
- Fix failing unit tests
- Remove unused files
- Update the SHA for Argo CD operator


chetan-rns:
@sbose78 Small patch to https://github.com/redhat-developer/gitops-operator/pull/24. Tested with the latest ArgoCD operator SHA and everything worked fine for me. Once this lands we can merge https://github.com/redhat-developer/gitops-operator/pull/24

sbose78 (approved):

PR #27, opened 2021-01-07
Serve kam CLI 

sbose78:
Associated with https://github.com/redhat-developer/kam/pull/139

Kam is served from https://kam-openshift-pipelines-app-delivery.apps.dev-svc-4.7-010701.devcluster.openshift.com/kam/

image


pkg/controller/gitopsservice/gitopsservice_controller_test.go
```diff
@@ -26,4 +26,26 @@ func TestImageFromEnvVariable(t *testing.T) {
 			t.Errorf("Image mismatch: got %s, want %s", got, backendImage)
 		}
 	})
+
+	t.Run("Kam Image present as env variable", func(t *testing.T) {
+		image := "quay.io/org/test"
+		os.Setenv(kamImageEnvName, image)
+		defer os.Unsetenv(kamImageEnvName)
```
bigkevmcd, line 33: I think this should be `cliImageEnvName`?

The tests are erroring without it.

pkg/controller/gitopsservice/kam.go
```diff
@@ -0,0 +1,215 @@
...
+			Labels: map[string]string{
+				"app": cliName,
+			},
+		},
+		Spec: podSpec,
+	}
+
+	var replicas int32 = 1
+	deploymentSpec := appsv1.DeploymentSpec{
+		Replicas: &replicas,
```
bigkevmcd, line 59: You could use https://godoc.org/knative.dev/pkg/ptr#Int32 here?

pkg/controller/gitopsservice/kam.go
```diff
@@ -0,0 +1,215 @@
...
+				Port:       port,
+				Protocol:   corev1.ProtocolTCP,
+				TargetPort: intstr.FromInt(int(port)),
+			},
+			{
+				Name:       "tcp-8443",
+				Port:       portSSL,
+				Protocol:   corev1.ProtocolTCP,
+				TargetPort: intstr.FromInt(int(8443)),
+			},
```
bigkevmcd, line 91: Shouldn't this be using portSSL? (which might be better named portTLS)

pkg/controller/gitopsservice/kam.go
```diff
@@ -0,0 +1,215 @@
...
+		},
+		Selector: map[string]string{
+			"app": cliName,
+		},
+	}
+	svc := &corev1.Service{
+		ObjectMeta: objectMeta(cliName, serviceNamespace, func(o *metav1.ObjectMeta) {}),
```
bigkevmcd, line 98: This func isn't needed?

pkg/controller/gitopsservice/kam.go
```diff
@@ -0,0 +1,215 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	template := corev1.PodTemplateSpec{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels: map[string]string{
+				"app": cliName,
```
bigkevmcd, line 51: Should this perhaps be using `app.kubernetes.io/name`, same below (which would be needed if this is changed).

pkg/controller/gitopsservice/kam.go
```diff
@@ -0,0 +1,215 @@
...
+		Port: &routev1.RoutePort{
+			TargetPort: intstr.IntOrString{IntVal: portSSL},
+		},
+		TLS: &routev1.TLSConfig{
+			Termination:                   routev1.TLSTerminationPassthrough,
+			InsecureEdgeTerminationPolicy: routev1.InsecureEdgeTerminationPolicyNone,
+		},
+	}
+
+	routeObj := &routev1.Route{
```
bigkevmcd, line 119: This could be simplified to just `return &routev1.Route{ ...`

pkg/controller/gitopsservice/kam.go
```diff
@@ -0,0 +1,215 @@
...
+			Description: text,
+			DisplayName: text,
+		},
+	}
+}
+
+func (r *ReconcileGitopsService) reconcileCLI(cr *pipelinesv1alpha1.GitopsService, request reconcile.Request) (reconcile.Result, error) {
+
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling GitopsService")
```
bigkevmcd, line 148: This is probably confusing, as this same message is logged out twice?

bigkevmcd (changes requested):

pkg/controller/gitopsservice/gitopsservice_controller_test.go
sbose78: Good catch, didn't test after the last minute renaming.

pkg/controller/gitopsservice/kam.go
sbose78: Right, removed.

pkg/controller/gitopsservice/kam.go
sbose78: Done, thank you!

pkg/controller/gitopsservice/kam.go
sbose78: Agreed!

pkg/controller/gitopsservice/kam.go
sbose78: Ah that was a miss, fixed.

pkg/controller/gitopsservice/kam.go
sbose78: Fixed and renamed 🚌

pkg/controller/gitopsservice/kam.go
sbose78: I could, wondering what the benefit is though 🤔

bigkevmcd: Not a blocker.

sbose78:
Thank you for your detailed review @bigkevmcd , I've addressed everything ( except the alternative to int32 ). Could you do another round of review?

bigkevmcd (approved):
This looks ok to me, I'm assuming the fact that it doesn't rereconcile anything in the event of changes is fine, it's not using the CR to generate any of the fields.

I'd also be inclined to retain a context, rather than using lots of different contexts, but that's stylistic.

And the test coverage isn't good.

sbose78:
>  I'm assuming the fact that it doesn't rereconcile anything in the event of changes is fine

If any of the owned resources change, a reconcile should happen. Let me know if I've missed something

> I'd also be inclined to retain a context,

Sounds fine to me, any suggestions on what a reasonable retention might look like? ( I see a ton of context.TODO() everywhere already, probably my bad.. ) 

> And the test coverage in't good.

I can improve!
Is there a specific scenario I should be adding?

sbose78:
> sbose78 dismissed bigkevmcd’s stale review via b1e9f80 33 seconds ago

I didn't do this, the bot did ;)

* Added ownerRef to the `ConsoleCLIDownload` CR. Yes, the objects don't get updated on reconcile. The loop only ensures they are present. ( we can improve this later )
* Took a look again, yes, I will add more tests, the existing tests need to validate the contents of the resources better.

pkg/controller/gitopsservice/kam.go
```diff
@@ -0,0 +1,212 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const cliName = "kam"
+const cliLongName = "GitOps Application Manager"
+const cliImage = "quay.io/redhat-developer/kam:1d0ef5"
```
wtam2018, line 24:
```suggestion
const cliImage = "registry.redhat.io/openshift-gitops-1-tech-preview/kam-delivery-rhel8:1.0-tech-preview"
```

wtam2018:
@sbose78 please resolve conflict file.

PR #29, opened 2021-01-18
Setup ArgoCD out of the box in the "openshift-gitops" 

sbose78:
Setup ArgoCD out of the box in the "openshift-gitops" namespace and enable it to use the cluster monitoring stack on OpenShift.

*Pending* 
~* Create a `ServiceMonitor`~ Can be done in a separate PR.
~* TESTS!~ Done, see comment below.


bigkevmcd:
Will this setup cross-namespace argocd-deployments correctly?

sbose78:
> Will this setup cross-namespace argocd-deployments correctly?

Do you mean the one like this?

```
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: argo-applier
  namespace: sbose-argocd-apps ### where my apps go
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: admin
subjects:
- kind: ServiceAccount
  name: argocd-application-controller
  namespace: openshift-pipelines-app-delivery   ### where argocd control plane is installed
```

pkg/controller/gitopsservice/argocd.go
```diff
@@ -0,0 +1,45 @@
+package gitopsservice
+
+import (
+	argoapp "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"gopkg.in/yaml.v2"
```
bigkevmcd, line 7: This should probably use `sigs.k8s.io/yaml` which has some subtle changes _on top_ of go yaml.

pkg/controller/gitopsservice/argocd.go
```diff
@@ -0,0 +1,45 @@
...
+func argoCDCR(name, ns string) (*argoapp.ArgoCD, error) {
+	b, err := yaml.Marshal([]resource{
+		{
+			APIGroups: []string{"tekton.dev"},
+			Kinds:     []string{"TaskRun", "PipelineRun"},
+			Clusters:  []string{"*"},
+		},
+	})
+	if err != nil {
+		// could use fmt.Errorf( ) but wouldn't make much of a difference I suppose.
```
bigkevmcd, line 26: We could drop this comment?

pkg/controller/gitopsservice/gitopsservice_controller.go
```diff
@@ -285,21 +307,21 @@ func newBackendDeployment() *appsv1.Deployment {
 		Replicas: &replicas,
 		Selector: &metav1.LabelSelector{
 			MatchLabels: map[string]string{
-				"app": serviceName,
+				"app": name,
```
bigkevmcd, line 310: Is there a reason not to use `app.kubernetes.io/name` here?

(If not, you'll need to update the code below too).

pkg/controller/gitopsservice/gitopsservice_controller.go
```diff
@@ -160,8 +162,28 @@ func (r *ReconcileGitopsService) Reconcile(request reconcile.Request) (reconcile
...
+		return reconcile.Result{}, err
+	}
+
+	// Set GitopsService instance as the owner and controller
+	if err := controllerutil.SetControllerReference(instance, defaultArgoCDInstance, r.scheme); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	existingArgoCD := &argoapp.ArgoCD{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: defaultArgoCDInstance.Name, Namespace: defaultArgoCDInstance.Namespace}, existingArgoCD)
```
bigkevmcd, line 176: There are no tests for this code.

bigkevmcd (changes requested):

pkg/controller/gitopsservice/gitopsservice_controller.go
sbose78: Yes, wanted to let folks know about the approach before writing tests. Will add them today :)

siamaksade:
@sbose78 would it make sense to call the namespace `openshift-gitops`? especially if the Argo CD instance would be used for cluster config management

christianh814:
> @sbose78 would it make sense to call the namespace `openshift-gitops`? especially if the Argo CD instance would be used for cluster config management

+1 ..this is consistent with other OpenShift Operators. They all start with `openshift-`

sbose78:
Fair points, we'll need to coordinate the UI changes accordingly. Moved out the namespace change to a different PR.

pkg/controller/gitopsservice/gitopsservice_controller.go
sbose78: 8c691fd thank you, fixed.

sbose78:
I've confirmed, we now have e2e tests that :

* Validate that ArgoCD CR is created in `openshift-gitops` namespace.
* Validate that ConsoleLink CR is created based on the route url of the argocd instance. ( needs https://github.com/argoproj-labs/argocd-operator/pull/233 merged and imported to pass )
* Validate that a human is allowed to modify the ArgoCD CR without it being overwritten by the controller.

cmd/manager/main.go
```diff
@@ -142,6 +143,7 @@ func main() {
 	registerComponentOrExit(mgr, operatorsv1alpha1.AddToScheme)
 	registerComponentOrExit(mgr, argoapi.AddToScheme)
 	registerComponentOrExit(mgr, configv1.AddToScheme)
+	registerComponentOrExit(mgr, monitoringv1.AddToScheme)
```
sbose78, line 146: This is needed for the operator to start up!

PR #34, opened 2021-01-25
Fix operator name and update icon

wtam2018:


deploy/olm-catalog/gitops-operator/manifests/gitops-operator.clusterserviceversion.yaml
```diff
@@ -96,9 +96,9 @@ spec:
         name: ''
         version: v1
   description: OpenShift GitOps is a layered product on top of OpenShift that enables teams to adopt GitOps principles for managing cluster configurations and automating secure and repeatable application delivery across hybrid multi-cluster Kubernetes environments. OpenShift GitOps is built around Argo CD as the core upstream project and assists customers to establish an end-to-end application delivery workflow on GitOps principles.
-  displayName: OpenShift Gitops
+  displayName: OpenShift GitOps Operator
```
sbose78, line 99: We wish to call it "OpenShift GitOps".

sbose78 (changes requested):

deploy/olm-catalog/gitops-operator/manifests/gitops-operator.clusterserviceversion.yaml
wtam2018: Oh.  It should be called "Red Hat OpenShift GitOps" per @siamaksade .    I will change it.

sbose78:
Thank you!

/lgtm
/approve

PR #39, opened 2021-01-29
env var for cluster config in deployment manifest

sbose78:
Use https://github.com/argoproj-labs/argocd-operator/blob/8aa14c03df16b634d6f01ac309f727f8645a6c4e/pkg/reconciler/openshift/openshift.go#L21


chetan-rns:
@sbose78 Since we have a single bundle, is the default value `openshift-gitops` valid for a 4.6 cluster?

sbose78:
If you are handling 4.6 separately, I would say No :) this is valid for 4.7 only.

sbose78:
I'll rescind what I said, and keep things simple. Let this be valid for all versions of openshift. That means, 4.6 will get a "app delivery" ArgoCD, that's fine in my opinion.

bigkevmcd (approved):

PR #42, opened 2021-02-01
Disable dex by default

chetan-rns:
Set `DISABLE_DEX` environment variable to true in the operator deployment. This prevents the operator controllers from deploying dex by default


bigkevmcd (approved):
Simple solution, thank you!

absynth76:
Could you please explain your choice about disabling dex ?

iam-veeramalla:
> Could you please explain your choice about disabling dex ?

@absynth76 Sorry but I did not understand the question. You can disable dex by setting DISABLE_DEX environment variable to true in the operator deployment

absynth76:
@iam-veeramalla  I would like to understand why redhat decided to disable dex by default.

iam-veeramalla:
> @iam-veeramalla I would like to understand why redhat decided to disable dex by default.

@absynth76 This was the initial decision as RedHat supports Keycloak but due to the number of requests we have decided to change this behavior. The v1.3 release planned for mid of OCT will have DEX installed and configured by default. No more manual installation or changes needed. 

User has to only update the RBAC as per their requirements. Please refer https://github.com/redhat-developer/gitops-operator/pull/204 for more details.

PR #47, opened 2021-02-03
Use Red Hat provided redis image as default

shubhamagarwal19:
This is in accordance to PR https://github.com/argoproj-labs/argocd-operator/pull/238.

GitOps operator by default should use a Red Hat provided redis image.

### DIFF
--- a/deploy/olm-catalog/gitops-operator/manifests/gitops-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/gitops-operator/manifests/gitops-operator.clusterserviceversion.yaml
@@ -160,6 +160,8 @@ spec:
                   value: gitops-operator
                 - name: DISABLE_DEX
                   value: "true"
+                - name: ARGOCD_REDIS_IMAGE
+                  value: "registry.redhat.io/rhel8/redis-5"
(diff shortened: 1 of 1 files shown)


bigkevmcd (dismissed):

chetan-rns:
Should we allow reading this image from an env variable for disconnected environments?
Example: https://github.com/redhat-developer/gitops-operator/blob/d473346f071cc38b54402bf402c863aabd7fcd38/pkg/controller/gitopsservice/kam.go#L29

Unit test for the above example can be found here https://github.com/redhat-developer/gitops-operator/blob/d473346f071cc38b54402bf402c863aabd7fcd38/pkg/controller/gitopsservice/gitopsservice_controller_test.go#L23

wtam2018:
@shubhamagarwal19  thanks for raising the PR.  Setting the redis image should be done by setting the environment variable in CSV. 
```
 name: ARGOCD_REDIS_IMAGE
 value: registry.redhat.io/rhel8/redis-5 
```

shubhamagarwal19:
@wtam2018, thanks for the comment. Updated the CSV with `ARGOCD_REDIS_IMAGE` ENV!

PR #51, opened 2021-02-08
Rename "Application Stages" to "OpenShift GitOps"

sbose78:
The ConsoleLink Header doesn't make much sense if called "Application Stages" anymore since the ArgoCD instance is also intended to be used for cluster configuration.


wtam2018:
/hold
Let's get TP out first

PR #52, opened 2021-02-11
pick up upstream fixes for dex reconcile

shubhamagarwal19:
Picks up the following fixes from upstream ArgoCD Operator - 

- https://github.com/argoproj-labs/argocd-operator/pull/243
- https://github.com/argoproj-labs/argocd-operator/pull/243

Issue:
The issue that this PR addresses is that the operator was creating Role, Role-Binding and Service-Account for Dex even when the service was disabled for OOTB ArgoCD instance. 

Tests: 
Performed manual test by deploying the GitOps Operator with above fixes and verified that specified resources were not created for the OOTB ArgoCD Instance anymore.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/redhat-developer/gitops-operator
 go 1.13
 
 require (
-	github.com/argoproj-labs/argocd-operator v0.0.15-0.20210205153727-cd8ed15cc61b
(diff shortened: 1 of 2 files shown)


wtam2018:
/hold
let's get TP out first

sbose78:
William, agreed. If this isn't an urgent / mandatory fix, this can probably wait.

PR #53, opened 2021-02-12
Fix e2e test for kam CLI link

chetan-rns:
Fixes:
* Register scheme for ConsoleCLIDownload
* Append `/kam/` to route URL so it matches the CLI download link


test/e2e/kam_test.go
```diff
@@ -38,12 +40,13 @@ func validateKamService(t *testing.T) {
 
 	// check ConsoleLink
```
jopit, line 41: The comment is now incorrect, it should say something like // check CLI download link

sbose78:
My bad that this slipped through the cracks, thank you for the fix!

PR #62, opened 2021-02-25
GITOPS-547: Auth Integration with RedHat SSO for Openshift Login

iam-veeramalla:

**What type of PR is this?**
> /kind enhancement

Enables SSO option for ArgoCD using RHSSO as Identity Broker and Openshift V4 as Identity Provider.  This is achieved by installing the below RHSSO resources and OpenShift OAuth Client.
1. Keycloak 
2. Keycloak Realm
3. Keycloak Client

However, apart from installing the above mentioned resources it also configures these resources to avoid any manual intervention in setting up SSO. 

Once the PR is merged the Users should be able to login into ArgoCD using their Openshift login. This PR does not create or provide steps on how to create openshift users.

This story does not need any documentation.

GITOPS-549 and GITOPS-662 (not the installation of operator , will be taken care in a different PR)

Fixes #?
GITOPS-549 and GITOPS-662 (not the installation of operator , will be taken care in a different PR)

1. Create "openshift-gitops" namespace
2.  Install RHSSO operator in "openshift-gitops" namespace
3. Install operator-sdk locally (recommended version - 0.17.0)
4. Run the below command 
`operator-sdk run --local --watch-namespace "openshift-gitops"`

login to openshift and you should see `login with keycloak`. Please follow the document for testing. 
```
https://github.com/dewan-ahmed/gitops-contents/blob/main/blogs/2021-02-16-openshift-sso-argocd.md
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -42,6 +42,8 @@ import (
 	_ "github.com/argoproj-labs/argocd-operator/pkg/reconciler/openshift"
 	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 
+	keycloakv1alpha1 "github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1"
+	oauthv1 "github.com/openshift/api/oauth/v1"
 	routev1 "github.com/openshift/api/route/v1"
 )
 
@@ -146,6 +148,8 @@ func main() {
 	registerComponentOrExit(mgr, argoapi.AddToScheme)
 	registerComponentOrExit(mgr, configv1.AddToScheme)
 	registerComponentOrExit(mgr, monitoringv1.AddToScheme)
+	registerComponentOrExit(mgr, keycloakv1alpha1.SchemeBuilder.AddToScheme)
+	registerComponentOrExit(mgr, oauthv1.AddToScheme)
 
 	// Add the Metrics Service
 	addMetrics(ctx, cfg)
--- a/deploy/olm-catalog/gitops-operator/manifests/gitops-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/gitops-operator/manifests/gitops-operator.clusterserviceversion.yaml
@@ -177,6 +177,12 @@ spec:
           - clusterrolebindings
           verbs:
           - '*'
+        - apiGroups:
+          - oauth.openshift.io
+          resources:
+          - oauthclients 
+          verbs:
+          - '*'
         serviceAccountName: gitops-operator
       permissions:
       - rules:
@@ -231,6 +237,29 @@ spec:
           - ingresses
           verbs:
           - '*'
+        - apiGroups:
+          - oauth.openshift.io
+          resources:
+          - oauthclient  
+          verbs:
+          - '*'
+        - apiGroups:
+          - keycloak.org
+          resources:
+          - keycloaks
+          - keycloaks/status
+          - keycloaks/finalizers
+          - keycloakrealms
+          - keycloakrealms/status
+          - keycloakrealms/finalizers
+          - keycloakclients
+          - keycloakclients/status
+          - keycloakclients/finalizers
+          verbs:
+          - create
+          - get
+          - list
+          - watch
         - apiGroups:
           - monitoring.coreos.com
           resources:
--- a/pkg/apis/pipelines/v1alpha1/gitopsservice_types.go
+++ b/pkg/apis/pipelines/v1alpha1/gitopsservice_types.go
@@ -6,6 +6,7 @@ import (
 
 // GitopsServiceSpec defines the desired state of GitopsService
 type GitopsServiceSpec struct {
+	EnableSSO bool `json:"enablesso,omitempty"`
 }
 
 // GitopsServiceStatus defines the observed state of GitopsService
--- a/pkg/controller/argocd/argocd.go
+++ b/pkg/controller/argocd/argocd.go
@@ -14,6 +14,15 @@ type resource struct {
 	Clusters  []string `json:"clusters"`
 }
 
+// OIDCConfig enables Integration with RHSSO
+type OIDCConfig struct {
+	Name           string   `json:"name"`
+	Issuer         string   `json:"issuer"`
+	ClientID       string   `json:"clientID"`
+	ClientSecret   string   `json:"clientSecret"`
+	RequestedScope []string `json:"requestedScopes"`
+}
+
 // NewCR returns an ArgoCD reference optimized for use in OpenShift
 // with Tekton
 func NewCR(name, ns string) (*argoapp.ArgoCD, error) {
--- a/pkg/controller/gitopsservice/gitopsservice_controller.go
+++ b/pkg/controller/gitopsservice/gitopsservice_controller.go
@@ -7,14 +7,17 @@ import (
 	"strings"
 
 	appsv1 "k8s.io/api/apps/v1"
+	"sigs.k8s.io/yaml"
 
-	argoapp "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
-	routev1 "github.com/openshift/api/route/v1"
-
-	pipelinesv1alpha1 "github.com/redhat-developer/gitops-operator/pkg/apis/pipelines/v1alpha1"
-	argocd "github.com/redhat-developer/gitops-operator/pkg/controller/argocd"
 	"github.com/redhat-developer/gitops-operator/pkg/controller/util"
 
+	argoapp "github.com/argoproj-labs/argocd-operator/pkg/apis/argoproj/v1alpha1"
+	keycloakv1alpha1 "github.com/keycloak/keycloak-operator/pkg/apis/keycloak/v1alpha1"
+	oauthv1 "github.com/openshift/api/oauth/v1"
+	routev1 "github.com/openshift/api/route/v1"
+	pipelinesv1alpha1 "github.com/redhat-developer/gitops-operator/pkg/apis/pipelines/v1alpha1"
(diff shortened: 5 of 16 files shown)


cmd/manager/main.go
```diff
@@ -92,6 +95,11 @@ func main() {
 
 	printVersion()
 
+	// Inspect cluster to verify availability of extra features
+	if err := util.InspectCluster(); err != nil {
+		log.Info("unable to inspect cluster")
```
wtam2018, line 100: print the error?

cmd/manager/main.go
```diff
@@ -144,6 +157,7 @@ func main() {
 	registerComponentOrExit(mgr, argoapi.AddToScheme)
 	registerComponentOrExit(mgr, configv1.AddToScheme)
 	registerComponentOrExit(mgr, monitoringv1.AddToScheme)
+	registerComponentOrExit(mgr, oauthv1.AddToScheme)
```
wtam2018, line 160: Move this inside the if-clause `util.IsKeycloakAPIAvailable()`?

pkg/apis/pipelines/v1alpha1/gitopsservice_types.go
```diff
@@ -6,6 +6,7 @@ import (
 
 // GitopsServiceSpec defines the desired state of GitopsService
 type GitopsServiceSpec struct {
+	SSO bool `json:"sso,omitempty"`
```
wtam2018, line 9: Consider changing the field name to more descriptive name like `KeycloakEnabled`, `SSOEnabled`.   SSO is actually a pretty board term.

deploy/olm-catalog/gitops-operator/manifests/gitops-operator.clusterserviceversion.yaml
```diff
@@ -176,6 +176,12 @@ spec:
           - clusterrolebindings
           verbs:
           - '*'
+        - apiGroups:
```
wtam2018, line 179: Please update role.yaml as well.

deploy/olm-catalog/gitops-operator/manifests/gitops-operator.clusterserviceversion.yaml
```diff
@@ -230,6 +236,29 @@ spec:
           - ingresses
           verbs:
           - '*'
+        - apiGroups:
+          - oauth.openshift.io
```
wtam2018, line 240: Please update role.yaml as well.

pkg/controller/argocd/argocd.go
```diff
@@ -14,6 +14,15 @@ type resource struct {
 	Clusters  []string `json:"clusters"`
 }
 
+// OIDCConfig enables Integration with RHSSO
+type OIDCConfig struct {
+	Name           string   `json:"name"`
+	Issuer         string   `json:"issuer"`
+	ClientID       string   `json:"clientID"`
+	ClientSecet    string   `json:"clientSecret"`
```
wtam2018, line 22: Please fix typo

pkg/controller/argocd/argocd.go
```diff
@@ -38,6 +47,7 @@ func NewCR(name, ns string) (*argoapp.ArgoCD, error) {
 		},
 		Spec: argoapp.ArgoCDSpec{
 			ResourceExclusions: string(b),
+			OIDCConfig:         "",
```
wtam2018, line 50: Do we have to put empty config when there is none (e.g. OIDC disabled)?

pkg/controller/gitopsservice/gitopsservice_controller.go
```diff
@@ -47,8 +50,22 @@ var (
 	serviceNamespace                  = "openshift-gitops"
 	depracatedServiceNamespace        = "openshift-pipelines-app-delivery"
 	clusterVersionName                = "version"
+	argoClientID                      = "openshift-gitops"
+	keycloakRouteName                 = "keycloak"
+	keycloakSecretName                = fmt.Sprintf("keycloak-client-secret-%s", rhsso.KeycloakArgoClient)
+	argoCDSecretName                  = "argocd-secret"
 )
 
+const gitopsIdentifier = "openshift-gitops"
```
wtam2018, line 59: Please seperate the OIDC/SSO stuff out from gitops service controller and create a new controller.

sbose78: Right.
Setting up of SSO needs to happen within the `GitOpsService` reconcile loop based on the `watch` on `GitOpsService`

deploy/olm-catalog/gitops-operator/manifests/gitops-operator.clusterserviceversion.yaml
iam-veeramalla: added the requested to role.yaml

cmd/manager/main.go
iam-veeramalla: moved the log to error level

pkg/controller/gitopsservice/gitopsservice_controller.go
wtam2018: @sbose78 Why?  Because sso is watching the ArgoCD route?   GitOpsService is very clustered  (with the argocd stuff)  The sso stuff does not seem to belong to GitOpsService.

sbose78 (changes requested):
Could you please add e2e tests?

pkg/controller/gitopsservice/gitopsservice_controller.go
sbose78: The sso service gets created if the GitOpsService CR has `.spec.sso=true`. In that case, aren't we watching the `GitOpsService` CR ?

pkg/controller/gitopsservice/gitopsservice_controller.go
```diff
@@ -264,7 +295,188 @@ func (r *ReconcileGitopsService) Reconcile(request reconcile.Request) (reconcile
...
+				err = r.client.Create(context.TODO(), defaultKeycloakRealmInstance)
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+			}
+
+			// Get Argocd Route URL and pass it to the Keycloak Client installation.
+			// Keycloak Client instance is created with default values out of the box. Admins or owners are allowed to modify the instance.
+			argoRoute := &routev1.Route{}
+			argoRouteRef := existingArgoRoute()
```
sbose78, line 361: Remind me, does the argocd controller create the route instantly, or does the route creation itself depend on something?

pkg/controller/gitopsservice/gitopsservice_controller.go
iam-veeramalla: By default, GitopsService CR has `.spec.sso=true`. gitopsservice_controller has a watch to `GitOpsService CR`

pkg/controller/gitopsservice/gitopsservice_controller.go
iam-veeramalla: it creates the route as soon as the instance gets created.

cmd/manager/main.go
iam-veeramalla: I think we don't need any condition to register the scheme, I have removed the if-clause and registered both keycloak and Oauth schemes.

pkg/apis/pipelines/v1alpha1/gitopsservice_types.go
iam-veeramalla: Changed the Name from SSO to EnableSSO

pkg/controller/argocd/argocd.go
iam-veeramalla: @wtam Can you point me to the typo ?

iam-veeramalla:
@sbose78  @wtam2018 

As Discussed, Made changes to the code so that it will not have dependency of RHSSO operator to be installed or Imported. 

This is how the User Workflow would like after the Change:
```
set enablesso=false by default
```
**To enable RHSSO Integration**
```
User installs RHSSO operator in openshift-gitops ns
User sets .spec.enablesso=true, gitopsservice watches for this change and creates the keycloak resources, Oauth resources and OIDC config.
```

pkg/apis/pipelines/v1alpha1/gitopsservice_types.go
wtam2018: That's probably fine.  My thought is that SSO is a board term.   SSO can mean OIDC/SAML/AD.   Our SSO here only means KeyCloak or OIDC.

pkg/controller/argocd/argocd.go
wtam2018: ClientSecet

iam-veeramalla: Sure .. Thanks @wtam2018  .. Also I see a minor issue with unit tests .. will fix it once I am back .

pkg/controller/gitopsservice/gitopsservice_controller.go
wtam2018: @sbose78  I think we should create a new resource.  GitOpsService was created for the purpose of managing backend service which does not have much to do with Argo CD SSO.   Also, it is an opportunity to start using a correct api group if we create a new resource.

pkg/controller/argocd/argocd.go
wtam2018: My other comment on this is that do any of these fields allow empty (omitempty)?

iam-veeramalla:
Closing the PR. This PR was initially created with an intent to support RHSSO only for the default instance in openshift-gitops operator. 

We came up with a better design to support/create and configure RHSSO for every argocd instance. This feature is developed and merged in upstream to support SSO using keycloak for every Argo CD instance.
https://github.com/argoproj-labs/argocd-operator/commit/f11b440c49df0ea3ead9d33d3484c7160b1397ef

PR #70, opened 2021-03-08
Rename OpenShift GitOps default Argo CD instance

ishitasequeira:
**What type of PR is this?**
kind code-refactoring

Rename OpenShift GitOps default Argo CD instance

https://issues.redhat.com/browse/GITOPS-659

Fixes #?

The steps to test the PR:

1. Log in to a cluster
2. oc create ns openshift-gitops
3. oc apply -f deploy/crds 
4. oc apply -f deploy/ -n openshift-gitops
5. oc delete deployment/(whatever the gitops operator deployment it creates is called) -n openshift-gitops
6. operator-sdk run --local --watch-namespace=""
7. oc project openshift-gitops
8.  oc get deployments 

```
NAME                                      READY   UP-TO-DATE   AVAILABLE   AGE
cluster                                   0/1     1            0           9s
kam                                       0/1     1            0           7s
openshift-gitops-application-controller   0/1     1            0           5s
openshift-gitops-dex-server               0/1     1            0           5s
openshift-gitops-redis                    0/1     1            0           5s
openshift-gitops-repo-server              0/1     1            0           5s
openshift-gitops-server                   0/1     1            0           4s
```

The steps we followed to test if the instance name is renamed :
1. Added the custom operator to the catalog
2. Installed the TP  operator
3. Created an Argo CD Application
4. Deleted the Argo CD instance
5. Deleted the operator
6. Installed the custom operator
7. logged in to the Argo CD UI. 
The App created by the previous instance was still present


pkg/controller/argocd/argocd_controller_test.go
```diff
@@ -19,7 +19,7 @@ import (
 )
 
 const (
-	argocdInstanceName = "argocd"
+	argocdInstanceName = "openshift-gitops"
```
chetan-rns, line 22: Instead of redefining the const in multiple files, we could use a common package to declare constants and reuse them everywhere, WDYT?  It'll be a single word change in the future

Something like https://github.com/argoproj/argo-cd/blob/master/common/common.go

chetan-rns (changes requested):
Thanks for the PR! We should update the service account name in kam as well

pkg/controller/argocdmetrics/argocd_metrics_controller_test.go
```diff
@@ -70,7 +72,7 @@ func TestReconcile_add_namespace_label(t *testing.T) {
 		namespace    string
 	}{
 		{
-			instanceName: "argocd-cluster",
+			instanceName: common.ArgoCDInstanceName,
```
wtam2018, line 75: In tests, we may not want to use constants (that are shared with production code) so we can catch typo bug or unexpected change in contant values.

wtam2018:
/lgtm
/approve
thanks @ishitasequeira !

sbose78:
/hold

We didn't promise smooth upgrades from 1.0.1 to 1.1 but wouldn't this change cause folks to have a second instance of ArgoCD ?

wtam2018:
We need to document in the release note.   A new Argo CD instance is installed after upgrade.  The old instance should be removed.  I am not sure if existing applications need to be re-created.  @jannfis

sbose78:
As discussed, let's ensure we have a good understanding of the situation we'll be putting 1.0.1 users in.

PR #71, opened 2021-03-09
Teardown ArgoCD instance in E2E tests

jannfis:
Signed-off-by: jannfis <jann@mistrust.net>

**What type of PR is this?**

> /kind enhancement

This adds tear-down of the `ArgoCD` instance we install during end-to-end testing. Currently, two subsequent runs of end-to-end testsuite on same cluster is not possible, because the `openshift-gitops` namespace won't be correctly deleted - it's stuck in terminating state due to a finalizer on the `ArgoCD` resource which cannot be satisfied, because the operator already is uninstalled when the resource should be deleted. Also, this test is good to have too.

Also, updated the e2e shell script to correctly constraint operator-sdk to pre-v1.0

Fixes #?

### DIFF
--- a/scripts/run_e2e_tests.sh
+++ b/scripts/run_e2e_tests.sh
@@ -5,8 +5,7 @@ set -x
 E2E_TEST_NS="gitops-test"
 E2E_TEST_DIR=./test/e2e
 ARGOCD_NS="openshift-gitops"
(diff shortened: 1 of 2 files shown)


wtam2018 (commented):
/lgtm
/approve

PR #85, opened 2021-03-16
Policy rules for cluster configuration

amitkrout:
**What type of PR is this?**
/kind enhancement

E2e test coverage

Fixes - Part of https://issues.redhat.com/browse/GITOPS-726

1. login OpenShift cluster
2. make test-e2e

### DIFF
--- a/test/e2e/gitops_service_test.go
+++ b/test/e2e/gitops_service_test.go
@@ -155,6 +155,13 @@ func validateArgoCDInstallation(t *testing.T) {
 	// check that this has not been overwritten
 	assert.Equal(t, existingArgoInstance.Spec.DisableAdmin, true)
 
+	// Check policy rules for cluster configuration
+	clusterRole := &rbacv1.ClusterRole{}
+	err = f.Client.Get(context.TODO(), types.NamespacedName{Name: "argocd-cluster-openshift-gitops-argocd-application-controller"}, clusterRole)
+	assertNoError(t, err)
+	expectedRules := policyRulesForClusterConfig()
+	assert.DeepEqual(t, clusterRole.Rules, expectedRules)
+
 }
 
 func assertNoError(t *testing.T, err error) {
@@ -236,4 +243,98 @@ func tearDownArgoCD(t *testing.T) {
 	err = e2eutil.WaitForDeletion(t, f.Client.Client, existingArgoInstance, retryInterval, timeout)
 	assertNoError(t, err)
 
+}
+
+func policyRulesForClusterConfig() []rbacv1.PolicyRule {
+	return []rbacv1.PolicyRule{
+		{
+			APIGroups: []string{
+				"operators.coreos.com",
+			},
+			Resources: []string{
+				"*",
+			},
+			Verbs: []string{
+				"*",
+			},
+		},
+		{
+			APIGroups: []string{
+				"operator.openshift.io",
+			},
+			Resources: []string{
+				"*",
+			},
+			Verbs: []string{
+				"*",
+			},
+		},
+		{
+			APIGroups: []string{
+				"user.openshift.io",
+			},
+			Resources: []string{
+				"*",
+			},
+			Verbs: []string{
+				"*",
+			},
+		},
+		{
+			APIGroups: []string{
+				"config.openshift.io",
+			},
+			Resources: []string{
+				"*",
+			},
+			Verbs: []string{
+				"*",
+			},
+		},
+		{
+			APIGroups: []string{
+				"console.openshift.io",
+			},
+			Resources: []string{
+				"*",
+			},
+			Verbs: []string{
+				"*",
+			},
+		},
+		{
+			APIGroups: []string{
+				"",
+			},
+			Resources: []string{
+				"namespaces",
+				"persistentvolumeclaims",
+				"persistentvolumes",
+				"configmaps",
+			},
+			Verbs: []string{
+				"*",
+			},
+		}, {
+			APIGroups: []string{
+				"rbac.authorization.k8s.io",
+			},
+			Resources: []string{
+				"*",
+			},
+			Verbs: []string{
+				"*",
+			},
+		}, {
+			APIGroups: []string{
+				"storage.k8s.io",
+			},
+			Resources: []string{
+				"*",
+			},
+			Verbs: []string{
+				"*",
+			},
+		},
+	}
 }


sbose78:
Looks great @amitkrout

amitkrout:
platform issue 
/retest

amitkrout:
platform issue
/retest

test/e2e/gitops_service_test.go
```diff
@@ -237,3 +244,97 @@ func tearDownArgoCD(t *testing.T) {
 	assertNoError(t, err)
 
 }
+
+func policyRulesForClusterConfig() []rbacv1.PolicyRule {
+	return []rbacv1.PolicyRule{
```
sbose78, line 249: For every resource type that's listed here, we need to have one test that creates an Argo `Application` CR and consumes manifests.

test/e2e/gitops_service_test.go
```diff
@@ -237,3 +244,97 @@ func tearDownArgoCD(t *testing.T) {
...
+			Resources: []string{
+				"*",
+			},
+			Verbs: []string{
+				"*",
+			},
+		},
+		{
+			APIGroups: []string{
+				"config.openshift.io",
```
sbose78, line 285: We should have a test that consumes https://github.com/openshift/openshift-gitops-examples/blob/master/argocd/builds/base/build.yaml#L2 in an Application CR

test/e2e/gitops_service_test.go
```diff
@@ -237,3 +244,97 @@ func tearDownArgoCD(t *testing.T) {
...
+			Resources: []string{
+				"*",
+			},
+			Verbs: []string{
+				"*",
+			},
+		},
+		{
+			APIGroups: []string{
+				"console.openshift.io",
```
sbose78, line 296: https://github.com/openshift/openshift-gitops-examples/blob/master/argocd/console/console.yaml

test/e2e/gitops_service_test.go
sbose78: And/or
https://github.com/openshift/openshift-gitops-examples/blob/master/argocd/image/image.yaml

test/e2e/gitops_service_test.go
```diff
@@ -237,3 +244,97 @@ func tearDownArgoCD(t *testing.T) {
...
+			Verbs: []string{
+				"*",
+			},
+		},
+		{
+			APIGroups: []string{
+				"",
+			},
+			Resources: []string{
+				"namespaces",
```
sbose78, line 310: https://github.com/siamaksade/openshift-gitops-getting-started/blob/main/cluster/namespace/spring-petclinic-ns.yaml

test/e2e/gitops_service_test.go
```diff
@@ -237,3 +244,97 @@ func tearDownArgoCD(t *testing.T) {
...
+				"persistentvolumeclaims",
+				"persistentvolumes",
+				"configmaps",
+			},
+			Verbs: []string{
+				"*",
+			},
+		}, {
+			APIGroups: []string{
+				"rbac.authorization.k8s.io",
```
sbose78, line 320: https://github.com/siamaksade/openshift-gitops-getting-started/blob/main/cluster/namespace/spring-petclinic-rolebinding.yaml

test/e2e/gitops_service_test.go
```diff
@@ -237,3 +244,97 @@ func tearDownArgoCD(t *testing.T) {
...
+		},
+		{
+			APIGroups: []string{
+				"",
+			},
+			Resources: []string{
+				"namespaces",
+				"persistentvolumeclaims",
+				"persistentvolumes",
+				"configmaps",
```
sbose78, line 313: https://github.com/openshift/openshift-gitops-examples/blob/master/argocd/scheduler/scheduler-config.yaml

test/e2e/gitops_service_test.go
```diff
@@ -237,3 +244,97 @@ func tearDownArgoCD(t *testing.T) {
 	assertNoError(t, err)
 
 }
+
+func policyRulesForClusterConfig() []rbacv1.PolicyRule {
+	return []rbacv1.PolicyRule{
+		{
+			APIGroups: []string{
+				"operators.coreos.com",
```
sbose78, line 252: https://github.com/sbose78/gitops-samples/blob/main/admin/logging/operator.yaml

sbose78 (commented):
These may go in, in a different PR. The goal is to have our e2e tests comprehensively validate if the rbac rules are not only present, but are being consumed effectively by ArgoCD.

amitkrout:
@sbose78 Thanks for a quick feedback. I am looking into the review comments

sbose78:
https://issues.redhat.com/browse/GITOPS-757

Should create at least one Application CR which references a helm chart.

test/e2e/gitops_service_test.go
```diff
@@ -237,3 +244,97 @@ func tearDownArgoCD(t *testing.T) {
 	assertNoError(t, err)
 
 }
+
+func policyRulesForClusterConfig() []rbacv1.PolicyRule {
```
shubhamagarwal19, line 248: @amitkrout, the policy rules defined here are older than the ones specified in ArgoCD Operator OpenShift hooks. 
I don't know if it's the best way to incorporate the policy rules in e2e tests. Maybe we can export the function (https://github.com/argoproj-labs/argocd-operator/blob/master/pkg/reconciler/openshift/openshift.go#L134) in ArgoCD Operator which could be utilized directly by the e2e tests so we don't need to maintain this at both places. 

Although the best way would be to move the OpenShift specific logic in ArgoCD operator to GitOps Opeartor

sbose78: Ummm... We shouldn't be using an exported function - if we do so this test will always pass. 

If policy rules change when a new version of argocd operator controller is vendor'd in, I would expect this test to fail and then explicitly fixed.

What do you think?

shubhamagarwal19: The function only returns the intended policy rules that should be specified in the Role, the actual test case is to verify the policy rules specified in the created Role with the intended policy rules, so I think it is ok to use the exported function to retrieve them. 
Also, the fix for it would be to copy the exact same rules from the ArgoCD Operator version imported as a vendor. 
So I was thinking, to avoid us the pain of updating the e2e tests every time we update the policy rules in ArgoCD operator, we can directly call the function.

amitkrout:
This pr will be worked on part of epic - https://issues.redhat.com/browse/GITOPS-715. All review comments will be addressed in the epic story.
/close

PR #97, opened 2021-03-25
Update new image registry path

amitkrout:
What type of PR is this?
/kind failing-ci

What does this PR do / why we need it:
To fix CI failure, Updating new test container registry path

Which issue(s) this PR fixes:

Fixes - CI failure due to deprecated test container registry

How to test changes / Special notes to the reviewer:
CI should pass


amitkrout:
CI will still fail because openshift CI job config still uses the master branch docker file which has the deprecated image registry path. So merging it forcefully cc_ @wtam2018

PR #110, opened 2021-04-22
add service account with required permissions for backend service

shubhamagarwal19:
**What type of PR is this?**
/kind enhancement

This PR creates a Cluster Role and Cluster Role Binding which provides the permissions to list/get/watch Application object in argoproj.io api and associate it with the Service Account for Gitops Backend Service (cluster). 

Creates SA/Cluster_Role/Cluster_Role_Binding with the name : "gitops-backend-service-cluster"

TODO: 

- [x] write unit tests

Fixes #?

- Deploy the latest GitOps operator with these changes.
- Check deployement yaml for GitOps Backend Service, it should have a `serviceAccount/serviceAccountName = gitops-backend-service-cluster`
- Check if SA/ClusterRole/ClusterRoleBinding exists with the above-provided name.
- Verify if the permissions specified in the cluster role allows the gitOps backend service to `GET/LIST/WATCH` Applications resource in all namespaces.

### DIFF
--- a/pkg/controller/gitopsservice/gitopsservice_controller.go
+++ b/pkg/controller/gitopsservice/gitopsservice_controller.go
@@ -50,7 +50,7 @@ var (
 )
 
 const (
-	gitopsBackendPrefix = "gitops-backend-service-"
+	gitopsServicePrefix = "gitops-service-"
 )
 
 // Add creates a new GitopsService Controller and adds it to the Manager. The Manager will set fields on the Controller
@@ -119,6 +119,30 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	}
 
 	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &pipelinesv1alpha1.GitopsService{},
+	})
+	if err != nil {
+		return err
+	}
+
+	err = c.Watch(&source.Kind{Type: &corev1.ServiceAccount{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &pipelinesv1alpha1.GitopsService{},
+	})
+	if err != nil {
+		return err
+	}
+
+	err = c.Watch(&source.Kind{Type: &rbacv1.ClusterRole{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &pipelinesv1alpha1.GitopsService{},
+	})
+	if err != nil {
+		return err
+	}
+
+	err = c.Watch(&source.Kind{Type: &rbacv1.ClusterRoleBinding{}}, &handler.EnqueueRequestForOwner{
 		IsController: true,
 		OwnerType:    &pipelinesv1alpha1.GitopsService{},
 	})
@@ -254,18 +278,18 @@ func (r *ReconcileGitopsService) Reconcile(request reconcile.Request) (reconcile
 	}
 
 	// Define Cluster Role Binding for backend service
-	cluserRoleBindingObj := newClusterRoleBinding(serviceNamespacedName)
+	clusterRoleBinding := newClusterRoleBinding(serviceNamespacedName)
 
 	// Set GitopsService instance as the owner and controller
-	if err := controllerutil.SetControllerReference(instance, cluserRoleBindingObj, r.scheme); err != nil {
+	if err := controllerutil.SetControllerReference(instance, clusterRoleBinding, r.scheme); err != nil {
 		return reconcile.Result{}, err
 	}
 
 	existingClusterRoleBinding := &rbacv1.ClusterRoleBinding{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: cluserRoleBindingObj.Name}, existingClusterRoleBinding)
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: clusterRoleBinding.Name}, existingClusterRoleBinding)
 	if err != nil && errors.IsNotFound(err) {
-		reqLogger.Info("Creating a new Cluster Role Binding", "Name", cluserRoleBindingObj.Name)
-		err = r.client.Create(context.TODO(), cluserRoleBindingObj)
+		reqLogger.Info("Creating a new Cluster Role Binding", "Name", clusterRoleBinding.Name)
+		err = r.client.Create(context.TODO(), clusterRoleBinding)
 		if err != nil {
 			return reconcile.Result{}, err
 		}
@@ -392,7 +416,7 @@ func newBackendDeployment(ns types.NamespacedName) *appsv1.Deployment {
 				},
 			},
 		},
-		ServiceAccountName: gitopsBackendPrefix + ns.Name,
+		ServiceAccountName: gitopsServicePrefix + ns.Name,
 	}
 
 	template := corev1.PodTemplateSpec{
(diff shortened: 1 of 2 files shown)


amitkrout:
> How to test changes / Special notes to the reviewer:

@shubhamagarwal19 Can you please add steps for how to verify it. The verification would be helpful to write integration/e2e test

shubhamagarwal19:
> > How to test changes / Special notes to the reviewer:
> 
> @shubhamagarwal19 Can you please add steps for how to verify it. The verification would be helpful to write integration/e2e test

updated !!

pkg/controller/gitopsservice/gitopsservice_controller.go
```diff
@@ -213,6 +217,60 @@ func (r *ReconcileGitopsService) Reconcile(request reconcile.Request) (reconcile
 		}
 	}
 
+	// Define Service account for backend Service
+	serviceAccountObj := newServiceAccount(serviceNamespacedName)
```
chetan-rns, line 221: New resources are not watched by the operator. It didn't reconcile back when the resources were updated/deleted.

chetan-rns (changes requested):

pkg/controller/gitopsservice/gitopsservice_controller.go
```diff
@@ -433,3 +492,59 @@ func newGitopsService() *pipelinesv1alpha1.GitopsService {
...
+
+func newClusterRoleBinding(meta types.NamespacedName) *rbacv1.ClusterRoleBinding {
+	return &rbacv1.ClusterRoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: gitopsBackendPrefix + meta.Name,
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      rbacv1.ServiceAccountKind,
+				Name:      meta.Name,
```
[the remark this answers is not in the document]
shubhamagarwal19: sorry, had the changes in my local forgot to push !! Thanks !

pkg/controller/gitopsservice/gitopsservice_controller.go
```diff
@@ -213,6 +217,60 @@ func (r *ReconcileGitopsService) Reconcile(request reconcile.Request) (reconcile
...
+	if err != nil && errors.IsNotFound(err) {
+		reqLogger.Info("Creating a new Cluster Role", "Name", clusterRoleObj.Name)
+		err = r.client.Create(context.TODO(), clusterRoleObj)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	// Define Cluster Role Binding for backend service
+	cluserRoleBindingObj := newClusterRoleBinding(serviceNamespacedName)
```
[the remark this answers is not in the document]
shubhamagarwal19: yes, makes sense !! updated, thanks !

pkg/controller/gitopsservice/gitopsservice_controller.go
shubhamagarwal19: yeah, I have seen a similar behaviour for other resources created by the operator. if I manually update the gitops backend service deployment, it doesn't reconcile either !! Maybe we can open a bug and handle this in a separate PR?

@chetan-rns, What do you suggest ?

pkg/controller/gitopsservice/gitopsservice_controller.go
```diff
@@ -49,6 +49,10 @@ var (
 	clusterVersionName                = "version"
 )
 
+const (
+	gitopsBackendPrefix = "gitops-backend-service-"
```
wtam2018, line 53: `gitops-service-` is probably fine.

pkg/controller/gitopsservice/gitopsservice_controller.go
wtam2018: The new resource should be watched.   

It looks like backend deployment/service are watched.  
https://github.com/redhat-developer/gitops-operator/blob/master/pkg/controller/gitopsservice/gitopsservice_controller.go#L101-L115

chetan-rns: The events are filtered by this predicate. https://github.com/redhat-developer/gitops-operator/blob/697a419b67b30cc1fcb81bd0db716ce17c8262f6/pkg/controller/gitopsservice/gitopsservice_controller.go#L69

shubhamagarwal19: ohh probably !! 
I'll fix the watch part of it ! Thanks @chetan-rns @wtam2018

pkg/controller/gitopsservice/gitopsservice_controller.go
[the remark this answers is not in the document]
shubhamagarwal19: Actually it needs to be a Cluster Role Binding, since it trying to list Application Resources in all namespaces.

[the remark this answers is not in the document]
chetan-rns: The role will be cluster scoped. But can we bind this cluster role using a local rolebinding?

Ex: We bind `admin` cluster role using a local rolebinding in every namespace where Argo CD instance is created.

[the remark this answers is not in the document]
shubhamagarwal19: Here we want to retrieve applications for all namespaces and not just `openshift-gitops`. if we bind a cluster with a role binding, then the permissions defined in the cluster role are applicable for that particular namespace only. 
Like the example, we bind `admin` cluster role to a role binding in `openshift-gitops` namespace, it provides service account the permissions to be an admin for openshift-gitops ns.

chetan-rns (approved):
/lgtm

Tested the PR locally.
1. SA, cluster role and cluster role binding is created
2. Resources are recreated after deletion
3. New resources have the right owner reference

pkg/controller/gitopsservice/gitopsservice_controller.go
shubhamagarwal19: done !

pkg/controller/gitopsservice/gitopsservice_controller.go
```diff
@@ -213,6 +241,60 @@ func (r *ReconcileGitopsService) Reconcile(request reconcile.Request) (reconcile
...
+	// Define Service account for backend Service
+	serviceAccountObj := newServiceAccount(serviceNamespacedName)
+
+	// Set GitopsService instance as the owner and controller
+	if err := controllerutil.SetControllerReference(instance, serviceAccountObj, r.scheme); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	existingServiceAccount := &corev1.ServiceAccount{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: serviceAccountObj.Namespace, Name: serviceAccountObj.Name}, existingServiceAccount)
```
sbose78, line 253: @shubhamagarwal19  Before this PR, which service account did this run as ? Why can't we create [cluster]rolebindings on that one?

shubhamagarwal19: it was using the `default` serviceAccount which doesn't have any permissions ... so it now creates a ClusterRole and RoleBinding which lets it `GET/LIST/WATCH` `Application resource in argoproj.io api` for all namespaces.

sbose78: Thanks, any reason we shouldn't add the permissions to the `default` sa ?

shubhamagarwal19: I think `default` SA is created by the k8s itself for each namespace. So I don't think it would be a nice idea to alter that? Also, any resource hor which `serivceAccount` isn't specified explicitly, uses the `default` SA for that namespace. 
So it would then have the permissions it shouldn't have.

sbose78: > Also, any resource where serivceAccount isn't specified explicitly, uses the default SA for that namespace, so it would then have the permissions it shouldn't have.

Makes sense, thank you 👍 

Have we done due diligence to ensure smooth upgrades?

Issue #130, opened 2021-06-16
Replace time.Sleep() in test suite


amitkrout:
### Is your task related to a problem? Please describe.

Currently test suite contains `time.Sleep()` which is not a optimize solution to wait for a status. 

### Describe the solution you'd like

Replace `time.Sleep()` with some kind of polling logic.

### Additional context

https://github.com/redhat-developer/gitops-operator/blob/b34fd46d1f3d2bf6dcaed2d0d072fa4ad0507298/test/e2e/gitops_service_test.go#L158

jgwest:
This is a good pattern to use for this: https://github.com/argoproj-labs/applicationset/blob/2c62537a8e5a3d5aecad87b843870789b74bdf89/test/e2e/fixture/applicationsets/consequences.go#L24

Basically it's:
- Pass a function into Expect that returns `true` if the test condition has passed, `false` otherwise
- In Expect, keep calling the function until it returns true.
- If it DOESN'T return true in X minutes (for example, 3 minutes), then fail the test.

Then, you can replace your sleep calls with tests for the particular cluster resource state that you are waiting for (with a timeout for if it never succeeds).

iam-veeramalla:
Hi @amitkrout , 

 You can use the `wait` package of kubernetes API Machinery. 
 [https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait)

iam-veeramalla:
We still have some entries of time.sleep() in the operator e2e ginkgo tests. This can be addressed and replaced with the `wait` package as mentioned in my previous comment. 

However, we are no longer maintaining these tests. We have already migrated to kUTTL tests which will be integrated with the CI asap.
cc: @varshab1210 , let me know if you need any help in hooking the kUTTL tests to our CI.

iam-veeramalla:
@wtam2018 @varshab1210 I have added `wontfix` label to this issue as we are interested in kUTTL tests going ahead.

PR #135, opened 2021-06-21
e2e test for cluster config change - GITOPS-1003

ishitasequeira:
What type of PR is this?
/kind enhancement

What does this PR do / why we need it:
This PR add the following test case.

Create an argocd application CR that has source folder containing a manifest of an config map in openshift-config project
oc apply -f the argocd application CR
Verify that the openshift cluster-config resource is deployed. Verify that the argocd application is healthy and sync'ed.

Have you updated the necessary documentation?

 Documentation update is required by this PR.
 Documentation has been updated.
Which issue(s) this PR fixes:

Fixes - https://issues.redhat.com/browse/GITOPS-1003

Test acceptance criteria:

 Unit Test - NA
 E2E Test
How to test changes / Special notes to the reviewer:

### DIFF
--- a/test/e2e/gitops_service_test.go
+++ b/test/e2e/gitops_service_test.go
@@ -578,17 +578,16 @@ func validateClusterConfigChange(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-
+	existingConfigMap := &corev1.ConfigMap{}
 	err = wait.Poll(time.Second*1, time.Second*60, func() (bool, error) {
-		if helper.ApplicationHealthStatus("scheduler", "openshift-gitops"); err != nil {
+		if err := helper.ApplicationHealthStatus("policy-configmap", "openshift-gitops"); err != nil {
 			t.Log(err)
 			return false, nil
 		}
-		if helper.ApplicationSyncStatus("scheduler", "openshift-gitops"); err != nil {
+		if err := helper.ApplicationSyncStatus("policy-configmap", "openshift-gitops"); err != nil {
 			t.Log(err)
 			return false, nil
 		}
-		existingConfigMap := &corev1.ConfigMap{}
 		if err := f.Client.Get(context.TODO(), types.NamespacedName{Name: "policy-configmap", Namespace: "openshift-config"}, existingConfigMap); err != nil {
 			t.Log(err)
 			return false, nil
--- a/test/e2e/gitops_service_test.go
+++ b/test/e2e/gitops_service_test.go
@@ -562,24 +562,47 @@ func skipOperatorDeployment() bool {
 }
 
 func validateClusterConfigChange(t *testing.T) {
-	framework.AddToFrameworkScheme(corev1.AddToScheme, &corev1.ConfigMap{})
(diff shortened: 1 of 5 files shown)


test/e2e/gitops_service_test.go
```diff
@@ -301,3 +303,38 @@ func validateMachineConfigUpdates(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	cmd := exec.Command(ocPath, "apply", "-f", schedulerYAML)
+	err = cmd.Run()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	time.Sleep(5 * time.Second)
```
jgwest, line 325: We probably don't want to add more `time.Sleep(...)` calls if we can help it. See https://github.com/redhat-developer/gitops-operator/issues/130 for outline of alternatives. 

You should be able to do something like this (pseudocode):
```go

// import this:
//	"k8s.io/apimachinery/pkg/util/wait"


err := wait.Poll(time.Second*1, time.Second*60, func() (bool, error) {
	
	if helper.ApplicationHealthStatus("scheduler", "openshift-gitops"); err != nil {
		// print the error received
		return false, nil
	}

	if helper.ApplicationSyncStatus("scheduler", "openshift-gitops"); err != nil {
		// print the error received	
		return false, nil
	}
	
	existingConfigMap := &corev1.ConfigMap{}
	err = f.Client.Get(context.TODO(), types.NamespacedName{Name: "policy-configmap", Namespace: "openshift-config"}, existingConfigMap)
	if err != nil {
			// print the error received
			return false, nil
	}

		return true, nil
})
```

ishitasequeira: I thought Amit was going to handle that in a separate task. But I can surely update this test.

amitkrout:
@ishitasequeira Can you please rebase.

test/helper/application_status.go
```diff
@@ -7,6 +7,25 @@ import (
...
+// ProjectExists return true if the AppProject exists in the namespace,
+// false otherwise (with an error, if available).
+func ProjectExists(projectName string, namespace string) (bool, error) {
+	var stdout, stderr bytes.Buffer
+	ocPath, err := exec.LookPath("oc")
+	if err != nil {
+		return false, err
+	}
+
+	cmd := exec.Command(ocPath, "get", "appproject/"+projectName, "-n", namespace, "-o", "yaml")
```
amitkrout, line 19: I think arguments ```"-o", "yaml"``` are not needed

test/helper/application_status.go
```diff
@@ -7,6 +7,25 @@ import (
...
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	if err := cmd.Run(); err != nil {
+		return false, fmt.Errorf("oc command failed: %s%s", stdout.String(), stderr.String())
```
amitkrout, line 23: Can we make it more expressive like 
```return false, fmt.Errorf("oc command failed. Stdout: %s, Stderr: %s", stdout.String(), stderr.String())```

amitkrout (changes requested):

amitkrout:
looks good to me. However i seek someone's review from the team. ping @wtam2018 
/approve

PR #147, opened 2021-06-29
ApplicationHealthStatus and ApplicationSyncStatus are assuming that an empty health/sync string means it is healthy/synced

jgwest:

ApplicationHealthStatus and ApplicationSyncStatus are assuming that an empty health/sync string means it is healthy/synced.

These functions should only return nil if the application is Healty and Synced, respectively.

Also:
- Using a slightly more efficient jsonpath query
- `cmd.Stdout = &stdout` should be before `cmd.Run()`
- Added `export ARGOCD_CLUSTER_CONFIG_NAMESPACES=openshift-gitops` to mimic the environment of the OLM-based install
- Removed another of the `time.Sleep`s
- The `validateMachineConfigUpdates` test is failing due to https://github.com/redhat-developer/gitops-operator/issues/148 (it previously appeared to pass in the test framework, but that was because ApplicationHealthStatus was not properly checking the health)
    -  So I added a Skip to it, for now, until #148 is fixed.

### DIFF
--- a/scripts/run_e2e_tests.sh
+++ b/scripts/run_e2e_tests.sh
@@ -24,6 +24,8 @@ fi
 # Create a new namespace for e2e tests
 oc new-project $E2E_TEST_NS
 
+export ARGOCD_CLUSTER_CONFIG_NAMESPACES=openshift-gitops
+
 echo "Running e2e tests"
 ${OPERATOR_SDK} test local $E2E_TEST_DIR --operator-namespace $E2E_TEST_NS --watch-namespace "" --up-local --verbose
 
--- a/scripts/run_e2e_tests.sh
+++ b/scripts/run_e2e_tests.sh
@@ -27,7 +27,7 @@ oc new-project $E2E_TEST_NS
 export ARGOCD_CLUSTER_CONFIG_NAMESPACES=openshift-gitops
 
 echo "Running e2e tests"
-${OPERATOR_SDK} test local $E2E_TEST_DIR --operator-namespace $E2E_TEST_NS --watch-namespace "" --up-local --verbose -timeout 30 
+${OPERATOR_SDK} test local $E2E_TEST_DIR --operator-namespace $E2E_TEST_NS --watch-namespace "" --up-local --verbose --go-test-flags "-timeout 30m"
 
 Teststatus=$?
 
--- a/scripts/run_e2e_tests.sh
+++ b/scripts/run_e2e_tests.sh
@@ -27,7 +27,7 @@ oc new-project $E2E_TEST_NS
 export ARGOCD_CLUSTER_CONFIG_NAMESPACES=openshift-gitops
 
 echo "Running e2e tests"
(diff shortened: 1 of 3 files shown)


jgwest:
FYI @amitkrout

test/e2e/gitops_service_test.go
```diff
@@ -278,6 +278,13 @@ func tearDownArgoCD(t *testing.T) {
 }
 
 func validateMachineConfigUpdates(t *testing.T) {
+
+	// This test will fail automation until gitops-operator #148 is fixed.
+	// 'When GitOps operator is run locally (not installed via OLM), it does not correctly setup
+	// the 'argoproj.io' Role rules for the 'argocd-application-controller'
+	// (https://github.com/redhat-developer/gitops-operator/issues/148)
+	t.SkipNow()
```
amitkrout, line 286: @jgwest We have a test target https://github.com/redhat-developer/gitops-operator/blob/master/Makefile#L31 that runs the same test on gitops-operator installed via OLM. We can skip this test with a condition, for example
```
if os.Getenv("SKIP_OPERATOR_DEPLOYMENT") != "true" {
		t.SkipNow()
	}
```

amitkrout (changes requested):

scripts/run_e2e_tests.sh
```diff
@@ -24,8 +24,10 @@ fi
 # Create a new namespace for e2e tests
 oc new-project $E2E_TEST_NS
 
+export ARGOCD_CLUSTER_CONFIG_NAMESPACES=openshift-gitops
```
amitkrout, line 27: AFAIU the value of key `ARGOCD_CLUSTER_CONFIG_NAMESPACES` is already set to `openshift-gitops` in https://github.com/redhat-developer/gitops-operator/blob/e9e2db1cc42006e1b24d663d5ce5a66785f826b1/deploy/operator.yaml#L33. Do we need it again ?

jgwest: We do: when running `run_e2e_tests.sh` that operator.yaml isn't used, so it's necessary to set the environment variable in `run_e2e_tests.sh`.

test/e2e/gitops_service_test.go
jgwest: Done.

scripts/run_e2e_tests.sh
amitkrout: Make sense. Thanks!

amitkrout:
@jgwest Test fails. PTAL
```
 === RUN   TestGitOpsService/Validate_machine_config_updates
I0630 16:01:05.832997    8680 request.go:621] Throttling request took 8.542981442s, request: GET:https://api.ci-op-w74ssqdz-02c74.origin-ci-int-aws.dev.rhcloud.com:6443/apis/project.openshift.io/v1?timeout=32s
    gitops_service_test.go:310: application 'image' health is ''
    gitops_service_test.go:310: application 'image' health is ''
    gitops_service_test.go:310: application 'image' health is ''
    gitops_service_test.go:310: application 'image' health is '' 
```

Test Validate_machine_config_updates does not skip in CI.

Issue #148, opened 2021-06-29
When GitOps operator is run locally (not installed via OLM), it does not correctly setup the 'argoproj.io' Role rules for the 'argocd-application-controller'


jgwest:
When GitOps operator is run locally, it does not correctly setup the 'argoproj.io' Role rules for the 'argocd-application-controller' pod/service account.

#### When all of these are true:
- You run the GitOps operator locally, for example via `scripts/run_e2e_tests.sh` or with `make run-local`
- You run it on a cluster that has not previously had the GitOps operator installed via OLM (and likewise not Argo CD operator)

#### You will find:
- That the `argocd-application-controller` created within the `openshift-gitops` namespace is missing permissions to read/write argoproj.io CRs.
- That the following error occurs in the `argocd-application-controller` pod log:
`
time="2021-06-28T19:46:10Z" level=error msg="Error persisting normalized application spec: applications.argoproj.io \"(application name)\" is forbidden: User \"system:serviceaccount:openshift-gitops:openshift-gitops-argocd-application-
controller\" cannot patch resource \"applications\" in API group \"argoproj.io\" in the namespace \"openshift-gitops\"" application=(application name)
`

This is due to how [the role is configured](https://github.com/argoproj-labs/argocd-operator/blob/3ee1128ee5a857ecec6be3b76eb471e174ce3790/pkg/reconciler/openshift/openshift.go#L79), as of https://github.com/argoproj-labs/argocd-operator/pull/318.

When OLM installs an operator, it adds read/write access to the `admin` ClusterRole. However, if you don't use OLM to install the GitOps operator, then the `admin` cluster role will NOT contain the Argo CD CR permissions.  Thus the `admin` cluster role cannot be guaranteed to contain the necessary cluster CRDs.

Unfortunately, as of https://github.com/argoproj-labs/argocd-operator/pull/318, the `admin` ClusterRole is appended to the `openshift-gitops-argocd-application-controller` namespace Role. So if the `admin` ClusterRole is missing the Argo CD CRs (because the operator was not installed via OLM), then the openshift-gitops namespace will be missing permissions to write to Argo CD CRs.

TLDR: we can't rely on the `admin` ClusterRole to necessarily contain the rules we need, because it is written to by OLM, and not every CRD is installed via OLM.

### To reproduce:

1. Start with a clean cluster (one on which no one has previously installed GitOps operator):
    - To verify that you have a clean cluster, run the following command:
[...]

PR #152, opened 2021-07-07
e2e test to validate permissions granted by managed-by label

chetan-rns:
**What type of PR is this?**
> /kind enhancement

Adds an e2e test to validate upstream PR [#346](https://github.com/argoproj-labs/argocd-operator/pull/346). The test validates if an ArgoCD instance in the source namespace can be granted permissions to create resources in the target namespace by adding the `managed-by` label.

Fixes #?

**Test acceptance criteria**:

* [ ] Unit Test
* [x] E2E Test


wtam2018 (commented):
/lgtm
/approve

test/e2e/gitops_service_test.go
```diff
@@ -548,6 +549,97 @@ func validateNonDefaultArgocdNamespaceManagement(t *testing.T) {
...
+	// create an ArgoCD app and check if it can create resources in the target namespace
+	taxiAppCr := filepath.Join("test", "appcrs", "taxi_appcr.yaml")
+	ocPath, err := exec.LookPath("oc")
+	assertNoError(t, err)
+	cmd := exec.Command(ocPath, "apply", "-f", taxiAppCr)
+	err = cmd.Run()
+	assertNoError(t, err)
+
+	err = wait.Poll(time.Second*1, time.Second*60, func() (bool, error) {
+		if err := helper.ApplicationHealthStatus("taxi", sourceNS); err != nil {
```
jaideepr97, line 630: @chetan-rns Shouldn't we be deploying the application in the target namespace and checking for it there rather than the source namespace?

chetan-rns: Argo CD can only manage applications in the namespace the instance is running, which is the source namespace.

jaideepr97: Isn't the idea that through this feature you'd be able to deploy applications in other namespaces that would be managed by the argocd-instance through the label? 
IIRC that is also what Shubham did in his demo

chetan-rns: Every non-default Argo CD instance will have admin rights in its namespace( the namespace where it's installed). So an instance running in NS-A won't have enough permissions to create k8s resources in NS-B. We can extend the access by explicitly creating a role/binding in NS-B granting access to Argo CD. Shubham's PR automates this step by allowing users to add labels and the operator would take care of roles/bindings behind the scene. The application CR would still reside in NS-A where Argo CD is running. In this context, the application refers to the Argo CD application CR that deploys k8s resources to the target namespace NS-B.

test/e2e/gitops_service_test.go
```diff
@@ -548,6 +549,97 @@ func validateNonDefaultArgocdNamespaceManagement(t *testing.T) {
...
+	}
+	err = f.Client.Create(context.TODO(), argoCDInstanceObj, cleanupOptions)
+	assertNoError(t, err)
+
+	// create a target namespace to deploy resources
+	// allow argocd to create resources in the target namespace by adding managed-by label
+	targetNS := "target-ns"
+	targetNamespaceObj := &corev1.Namespace{
+		ObjectMeta: v1.ObjectMeta{
+			Name: sourceNS,
```
shubhamagarwal19, line 588: shouldn't this be `targetNS` ?

test/e2e/gitops_service_test.go
shubhamagarwal19: @jaideepr97, I think what Chetan's trying to say here is that the Application CR will reside in the `source-ns` while the application would be deployed in `target-ns` which is how the application deployment in ArgoCD works. So I think that should be fine !!

jaideepr97: Oh got it, thanks @chetan-rns @shubhamagarwal19

test/e2e/gitops_service_test.go
```diff
@@ -633,33 +741,65 @@ func validateClusterConfigChange(t *testing.T) {
 
 }
 
-func applyMissingPermissions(f *framework.Framework, ocPath string) error {
-
+func applyMissingPermissions(ctx *framework.Context, f *framework.Framework, name, namespace string) error {
```
chetan-rns, line 744: Had to refactor this function to create roles in other namespaces. Ref: [Issue #148 ](https://github.com/redhat-developer/gitops-operator/issues/148)

test/e2e/gitops_service_test.go
chetan-rns: Thanks! I'ts updated

shubhamagarwal19 (approved):
@chetan-rns Thanks for the PR! LGTM!!

PR #161, opened 2021-07-09
Improve helper namespace deletion

jgwest:
**What type of PR is this?**

/kind cleanup

**What does this PR do / why we need it**:

Improve the helper namespace deletion logic I added in a previous PR

It is possible for the `openshift-gitops` namespace to still have an active `ArgoCD` instance, at the end of the previous E2E test, which blocks deletion of the `openshift-gitops` namespace (due to the ArgoCD finalizer). 

Which looks like this:
```
I0709 12:50:50.199629   10291 request.go:621] Throttling request took 1.13356149s, request: GET:https://api.ci-op-dpljqn95-02c74.origin-ci-int-aws.dev.rhcloud.com:6443/apis/build.openshift.io/v1?timeout=32s
I0709 12:51:00.397791   10291 request.go:621] Throttling request took 11.331586527s, request: GET:https://api.ci-op-dpljqn95-02c74.origin-ci-int-aws.dev.rhcloud.com:6443/apis/coordination.k8s.io/v1beta1?timeout=32s
    application_status.go:133: Running ensureCleanSlate
    application_status.go:154: Namespace 'gitops-standalone-test' no longer exists
    application_status.go:158: Namespace 'openshift-gitops' still exists
    application_status.go:158: Namespace 'openshift-gitops' still exists
    application_status.go:158: Namespace 'openshift-gitops' still exists
    application_status.go:158: Namespace 'openshift-gitops' still exists
    application_status.go:158: Namespace 'openshift-gitops' still exists
    application_status.go:158: Namespace 'openshift-gitops' still exists
    application_status.go:158: Namespace 'openshift-gitops' still exists
    application_status.go:158: Namespace 'openshift-gitops' still exists
    (until it expires)
```

This PR patches the ArgoCD (in the namesapce) to remove the finalizer, as part of the namespace deletion.

**How to test changes / Special notes to the reviewer**:

PR #166, opened 2021-07-12
Increase timeout for a flaky e2e test

chetan-rns:
**What type of PR is this?**
> /kind failing-test

e2e test `validateGrantingPermissionsByLabel` is flaky because of a timeout issue. This PR increases the timeout duration

Fixes #?

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test

### DIFF
--- a/test/e2e/gitops_service_test.go
+++ b/test/e2e/gitops_service_test.go
@@ -587,7 +587,7 @@ func validateGrantingPermissionsByLabel(t *testing.T) {
 	err = cmd.Run()
 	assertNoError(t, err)
 
-	err = wait.Poll(time.Second*1, time.Second*180, func() (bool, error) {
+	err = wait.Poll(time.Second*1, time.Minute*5, func() (bool, error) {
 		if err := helper.ApplicationHealthStatus("nginx", sourceNS); err != nil {
 			t.Log(err)
 			return false, nil


test/e2e/gitops_service_test.go
```diff
@@ -587,7 +587,7 @@ func validateGrantingPermissionsByLabel(t *testing.T) {
 	err = cmd.Run()
 	assertNoError(t, err)
 
-	err = wait.Poll(time.Second*1, time.Second*180, func() (bool, error) {
+	err = wait.Poll(time.Second*1, time.Minute*5, func() (bool, error) {
```
iam-veeramalla, line 590: I dont think increase the timeout to 5 minutes is a right thing to do in this case. We should identify the actual point of intermittent or flaky behavior :)

iam-veeramalla (changes requested):

test/e2e/gitops_service_test.go
wtam2018: This is a wait poll timeout which means the longest it would wait.  Do we see any evidence that it can normally take more than 100 seconds?

chetan-rns: Sometimes it takes around 6-7 minutes to reach a healthy state. Not sure what's causing this delay. I see the timeout is now increased to 10 mins by PR #161

jgwest: Agreed that increasing it isn't the best solution: personally I investigated the issue and ran out of leads (for example was never able to reproduce on clusterbot cluster with e2e tests run in a loop), and increased it more out of frustration then good sense :upside_down_face:.

chetan-rns: I couldn't reproduce it manually as well. The app reached a healthy state seamlessly without any delay

wtam2018: /lgtm
/approve

wtam2018:
timeout has been increased by another PR #161

PR #179, opened 2021-07-21
fix: set resource requirements to containers incase of the operator upgrade

iam-veeramalla:

**What type of PR is this?**
> /kind bug

This PR addresses the following issue.

After upgrading to v1.2, Resource Quota has been configured in `openshift-gitops` namespace by the operator.   However, existing Deployments do not contain Resource limit/request.   When any one of the existing pods are deleted, new pods failed to start because resource limit/requests are not defined in the Deployments.`

This PR makes the following changes:
* When reconciling the default instance, we make sure the following workloads contain resource request/limit definitions.
** applicationset controller
** application controller
** dex
** grafana
** ha proxy
** redis
** repo server
* Reconcile resource limit/request changes of backend service
* Reconcile resource limit/request changes of kam delivery container

NA

**Test acceptance criteria**:
* [ ] Unit Test
* [ ] E2E Test

1. Install the operator v1.1.2
2. Upgrade to the new version of operator v1.2
3. Delete any pod, you will notice that a new pod will never move into running state as it is not created with resource requirements.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/redhat-developer/gitops-operator
 go 1.16
 
 require (
-	github.com/argoproj-labs/argocd-operator v0.0.16-0.20210716115333-5b13ae4379ba
+	github.com/argoproj-labs/argocd-operator v0.0.16-0.20210722130040-5406a672631a
 	github.com/bugsnag/bugsnag-go v1.5.3 // indirect
 	github.com/bugsnag/panicwrap v1.2.0 // indirect
 	github.com/coreos/prometheus-operator v0.40.0
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/redhat-developer/gitops-operator
 go 1.16
 
 require (
-	github.com/argoproj-labs/argocd-operator v0.0.16-0.20210722130040-5406a672631a
+	github.com/argoproj-labs/argocd-operator v0.0.16-0.20210722160114-5fe7ef0c459f
 	github.com/bugsnag/bugsnag-go v1.5.3 // indirect
 	github.com/bugsnag/panicwrap v1.2.0 // indirect
 	github.com/coreos/prometheus-operator v0.40.0
--- a/pkg/controller/gitopsservice/gitopsservice_controller.go
+++ b/pkg/controller/gitopsservice/gitopsservice_controller.go
@@ -368,8 +368,59 @@ func (r *ReconcileGitopsService) reconcileDefaultArgoCDInstance(instance *pipeli
 		} else {
 			return reconcile.Result{}, err
 		}
-	}
+	} else {
+		changed := false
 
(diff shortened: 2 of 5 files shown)


pkg/controller/gitopsservice/kam.go
```diff
@@ -163,7 +164,8 @@ func (r *ReconcileGitopsService) reconcileCLIServer(cr *pipelinesv1alpha1.Gitops
 	}
 
 	// Check if this Deployment already exists
-	err := r.client.Get(context.TODO(), types.NamespacedName{Name: deploymentObj.Name, Namespace: deploymentObj.Namespace}, &appsv1.Deployment{})
+	found := &appsv1.Deployment{}
```
wtam2018, line 167: I'd name ths variable as something like `existingDeployment`.  `found` sounds like a bool.

iam-veeramalla: Acknowledge

iam-veeramalla:
Dependency on https://github.com/argoproj-labs/argocd-operator/pull/376

pkg/controller/gitopsservice/gitopsservice_controller.go
```diff
@@ -493,6 +544,10 @@ func (r *ReconcileGitopsService) reconcileBackend(gitopsserviceNamespacedName ty
 				found.Spec.Template.Spec.Containers[0].Args = deploymentObj.Spec.Template.Spec.Containers[0].Args
 				changed = true
 			}
+			if found.Spec.Template.Spec.Containers[0].Resources.Requests == nil {
+				found.Spec.Template.Spec.Containers[0].Resources = deploymentObj.Spec.Template.Spec.Containers[0].Resources
+				changed = true
+			}
```
wtam2018, line 550: We probably should do something like this instead of checking for non nil `Requests` field.
```
if !reflect.DeepEqual(deploy.Spec.Template.Spec.Containers[0].Resources, deploymentObj.Spec.Template.Spec.Containers[0].Resources) {

```

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/redhat-developer/gitops-operator
 go 1.16
 
 require (
-	github.com/argoproj-labs/argocd-operator v0.0.16-0.20210716115333-5b13ae4379ba
+	github.com/argoproj-labs/argocd-operator v0.0.16-0.20210722130040-5406a672631a
```
wtam2018, line 6: Please update to pick up https://github.com/argoproj-labs/argocd-operator/pull/378 as well.

pkg/controller/gitopsservice/gitopsservice_controller.go
iam-veeramalla: acknowledge :)

go.mod
iam-veeramalla: Sure :)

Issue #208, opened 2021-08-31
Document steps to create, update and delete resource requests/limits for Argo CD workloads.


iam-veeramalla:
**Describe the bug**
Many users have reported the OOM issues on the `Argo CD` instance created by the `gitops-operator`. Most of these issues are on `repo-server` and `application-controller`. The issue can easily be fixed by updating the resource limits of the specific workload that failed.

The resource requests/limits can be updated for any or all the workloads by updating the Argo CD CR. However, there is no proper documentation which explains the steps to perform this.

Please create a well explained documentation to fix this issue.

tux-o-matic:
A solution to avoid OoM in some cases involves #211, especially to avoid having to set a really high GB memory limit just to deal with peak memory usage during application controller startup .

Issue #207, opened 2021-08-31
Increase the default resource memory requirement for Repo Server.


iam-veeramalla:
**Describe the bug**
From v1.2, All the Argo CD workloads in the default Instance namespace are deployed with resource requests/limits by default. The current resource requests and limits set on the repo server is:

```
Requests: v1.ResourceList{
	Memory: "256Mi",
        CPU:   "250m",
},
Limits: 
       Memory: "512Mi",
       CPU:    "1000m",
```
However, many users have reported OOM issues on the repo-server. Looking at the issues I think we would need to upgrade the memory from `512Mi` to `1024Mi`

**To Reproduce**
1. run the operator from the console or from the CLI using `operator-sdk run local`
2. Move to Workloads -> Deployment -> `openshift-gitops-repo-server`  and look at the memory limit that is set.

**Expected behavior**
OOM crash should be avoided.

**Screenshots**

**Additional context**

Issue #206, opened 2021-08-31
Remove resource quota set for the default Argo CD Instance namespace.


iam-veeramalla:
**Describe the bug**
v1.2 sets resource quota `openshift-gitops-compute-resources` for the default Argo CD instance namespace. This is created with the intention to restrict the namespace with resources and ensure all the workloads are deployed with resource requests and limits.

Users have noted many issues with this default behavior. 

1. `gitops-operator` is an openshift specific operator. So the resource quota is set with the scope "Not Terminating" to support the deployment of deployer pods(specific to deployment configs[DC]). Users have reported issue running this in combination with some machine learning pods which are of scope "Terminating".
2. Few users are using the default Argo CD instance for app delivery. In such cases, Argo CD instance may try to deploy the pods in `openshift-gitops` namespace which would lead to a failure if the pods does not have resource requests/limits set.
3. The current resource quota is not well tested and users might have to upscale and downscale the quota as per their requirements.

**Expected behavior**
Remove the resource quota on the default Argo CD instance namespace.

**Additional context**
NA

PR #214, opened 2021-09-08
fix: issues with resource quota and resource requirements (v1.2.1)

iam-veeramalla:

**What type of PR is this?**
> /kind bug
> /kind documentation

Documentation is updated as part of the PR https://github.com/argoproj-labs/argocd-operator/pull/412

 * Remove resource quota in the namespace openshift-gitops in the controller reconciler.  When the operator is upgraded, existing resource quota is removed.
 * Increase the default memory limit to 1024M in the repo server.  It only affects new installation.  Existing derault argocd instance workload is not affected.

Fixes #?
#206 #207 #208 

**Test acceptance criteria**:
* [x] Unit Test

1. run the operator locally using `operator-sdk run local`
2. verify that there is no resource quota on the `openshift-gitops` namespace.
3. verify that the repo server memory limit is increased from `512Mi` to `1024Mi`
4. verify that a well explained documentation is available for managing resource requirements of Argo CD workloads.

### DIFF
--- a/pkg/controller/gitopsservice/gitopsservice_controller.go
+++ b/pkg/controller/gitopsservice/gitopsservice_controller.go
@@ -316,37 +316,26 @@ func (r *ReconcileGitopsService) reconcileDefaultArgoCDInstance(instance *pipeli
 		} else {
 			return reconcile.Result{}, err
 		}
-	}
-
-	// Ensure compute resource quotas are set for the default ArgoCD namespace
-	resourceQuotaName := argocdNS.Name + "-compute-resources"
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: resourceQuotaName, Namespace: argocdNS.Name}, &corev1.ResourceQuota{})
-	if err != nil && errors.IsNotFound(err) {
-		reqLogger.Info("Setting compute resource quotas for ArgoCD namespace", "Name", argocdNS.Name)
-		resourceQuota := corev1.ResourceQuota{
-			TypeMeta: metav1.TypeMeta{
-				Kind:       "ResourceQuota",
-				APIVersion: "v1",
-			},
+	} else {
+		// Delete if resource quota is set on the default ArgoCD instance namespace.
+		// Fix for v1.2 - https://github.com/redhat-developer/gitops-operator/issues/206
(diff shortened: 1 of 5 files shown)


pkg/controller/gitopsservice/gitopsservice_controller.go
```diff
@@ -368,59 +357,8 @@ func (r *ReconcileGitopsService) reconcileDefaultArgoCDInstance(instance *pipeli
 		} else {
 			return reconcile.Result{}, err
 		}
-	} else {
-		changed := false
-
-		if existingArgoCD.Spec.ApplicationSet != nil {
```
wtam2018, line 374: Why are we deleting all these?  We are still setting resource limits on the workloads, correct?

iam-veeramalla: for upgrades to v1.2 , we forcefully added this logic to ensure all the resources have requests/limits as the default namespace has resourcequota. Since, we removed the resourcequota object from v1.2.1 on default namespace this forceful updation is not needed.

Moreever, we dont know how many applications is the instance managing with the previous versions. So, these defaults might give a breaking experience.

wtam2018: I thought we were removing quota in the namespace and leaving resource requests/limits (with correct defaults) in the workloads.   When the user want to enforce quota, he/she can do so by adding quota in the ArgoCD CR.

iam-veeramalla: For the fresh install yes I followed the same approach but for the upgrades I choose not to force any resource limits as we don't know the size of their applications. We can discuss on this for consensus

wtam2018: For upgrade existing argocd workloads with resource limits/resources, this PR would have no effect on them as we only increase the default memory limit for repo server.   The default will not be to  any existing workload that already has resource limit configured.

iam-veeramalla: Ack. Reverted the changes and tests.

iam-veeramalla: @wtam2018 PR is ready for review again.

PR #215, opened 2021-09-09
feat: add toggle to enable openshift infra node and tolerations

saumeya:
**What type of PR is this?**

 /kind enhancement

This PR enables our operator to run on the openshift infra node, it also allows the user to add tolerations for the gitops workloads and the default argocd instance

**Test acceptance criteria**:

* [x] Unit Test
* [x] E2E Test

To run the operator locally `make install` and `make run` , then add the following in the gitops service CR. 
The pods will go in Pending state if no node with `node-role.kubernetes.io/infra: ""` is available. Label any worker node and the pods will reschedule on infra node.
doc - https://docs.google.com/document/d/1rZR4LvjSooRN-8o0-Wx-qOuyU2HVmpx0LdlS8Ku7C9o/edit?usp=sharing
spec:
  infraNodeEnabled: true
  tolerations:
  -effect: NoSchedule
     key: infra
     value: reserved
  -effect: NoExecute
     key: infra
     value: reserved


wtam2018:
@saumeya please take a look at unit test failure.

api/v1alpha1/gitopsservice_types.go
```diff
@@ -17,13 +17,19 @@ limitations under the License.
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // GitopsServiceSpec defines the desired state of GitopsService
 type GitopsServiceSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
+
+	// InfraNodeEnabled will add infra NodeSelector to all the default workloads of gitops operator
+	InfraNodeEnabled bool `json:"infraNodeEnabled,omitempty"`
```
iam-veeramalla, line 30: Can we change this from `InfraNodeEnabled` to `RunOnInfra` ?

api/v1alpha1/gitopsservice_types.go
```diff
@@ -17,13 +17,19 @@ limitations under the License.
...
+	// Tolerations allow the pods to schedule onto nodes with matching taints
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
```
iam-veeramalla, line 32: If I am not wrong, the only choice this implementation provides is to deploy the default workloads on Infra Node right ? So can we change the `Tolerations` to `TolerationsForInfraNode` ?

config/crd/bases/pipelines.openshift.io_gitopsservices.yaml
```diff
@@ -35,6 +35,52 @@ spec:
             type: object
           spec:
             description: GitopsServiceSpec defines the desired state of GitopsService
+            properties:
+              infraNodeEnabled:
```
iam-veeramalla, line 39: If we agree to change the naming convention as suggested in the above comments. Please run `make manifests` and `make bundle` after the change.

iam-veeramalla (changes requested):
Please add e2e tests and unit tests to validate the behavior.

controllers/gitopsservice_controller.go
```diff
@@ -321,6 +321,41 @@ func (r *ReconcileGitopsService) reconcileDefaultArgoCDInstance(instance *pipeli
...
+	}
+	if len(instance.Spec.Tolerations) > 0 {
+		if defaultArgoCDInstance.Spec.NodePlacement == nil {
+			defaultArgoCDInstance.Spec.NodePlacement = &argoapp.ArgoCDNodePlacementSpec{
+				Tolerations: instance.Spec.Tolerations,
+			}
+		} else {
+			defaultArgoCDInstance.Spec.NodePlacement.Tolerations = instance.Spec.Tolerations
+		}
+	}
```
jaideepr97, line 358: @saumeya Looks like this block of code may have been repeated accidentally

api/v1alpha1/gitopsservice_types.go
saumeya: sure

api/v1alpha1/gitopsservice_types.go
saumeya: while the use case most of the time will be for infra node, tolerations can be used independently as well

config/crd/bases/pipelines.openshift.io_gitopsservices.yaml
saumeya: sure

controllers/gitopsservice_controller.go
saumeya: thanks a lot for noticing jaideep, must have happened with some rebase. i'll correct it

controllers/gitopsservice_controller.go
```diff
@@ -482,7 +521,14 @@ func (r *ReconcileGitopsService) reconcileBackend(gitopsserviceNamespacedName ty
 		if err := controllerutil.SetControllerReference(instance, deploymentObj, r.Scheme); err != nil {
 			return reconcile.Result{}, err
 		}
-
+		if instance.Spec.InfraNodeEnabled {
+			deploymentObj.Spec.Template.Spec.NodeSelector = map[string]string{
+				"node-role.kubernetes.io/infra": "",
```
jaideepr97, line 526: Since this label is repeating in many places, would it make sense to instead define it in `common/common.go` and reference it from there? Maybe more maintainable for the future

cc @iam-veeramalla

common/common.go
```diff
@@ -22,3 +22,10 @@ const (
 	// DisableDefaultInstallEnvVar is an env variable to disable the default instance
 	DisableDefaultInstallEnvVar = "DISABLE_DEFAULT_ARGOCD_INSTANCE"
 )
+
+// InfraNodeSelector is openshift label for infrastructure nodes since maps are not constants returning from a funciton
+func InfraNodeSelector() map[string]string {
+	return map[string]string{
+		"node-role.kubernetes.io/infra": "",
+	}
+}
```
jaideepr97, line 31: @saumeya We probably don't need a new function just for this, what I meant earlier was just a suggestion that we could define a new constant in this file,something like  `InfraNodeLabel = "node-role.kubernetes.io/infra" `

And use it wherever needed as ```common.InfraNodeLabel: "" ``` (such as [here](https://github.com/redhat-developer/gitops-operator/pull/215/files#diff-83cb198211fecd2e48f050de2a80a348f8525ade03f84535fbe925fb8dc3608aR328) )

The idea was just that in case the label changes in the future we can just make the change in this one place rather than having to change it 5-6 times if we hard code the string in multiple places

saumeya: oh that wont work, becasue nodeSelector is a map not a string

saumeya: got it Jaideep, but adding a string constant in common file won't decrease the size of code at all, i'll still have to instantiate a map everytime. Unless having a common function is incorrect for the purpose of common file, can this part remain?

common/common.go
```diff
@@ -22,3 +22,10 @@ const (
 	// DisableDefaultInstallEnvVar is an env variable to disable the default instance
 	DisableDefaultInstallEnvVar = "DISABLE_DEFAULT_ARGOCD_INSTANCE"
 )
+
+// InfraNodeSelector is openshift label for infrastructure nodes since maps are not constants returning from a funciton
```
iam-veeramalla, line 26: You dont have to explain the technical reasons for using a function in the comments. If you still think it is useful .. Please correct the typo "funciton" to "function".

controllers/gitopsservice_controller_test.go
```diff
@@ -443,6 +443,40 @@ func TestGetBackendNamespace(t *testing.T) {
 	})
 }
 
+func TestReconcile_InfrastructureNode(t *testing.T) {
```
iam-veeramalla, line 446: 👍

controllers/gitopsservice_controller_test.go
```diff
@@ -473,3 +507,20 @@ func assertNoError(t *testing.T, err error) {
 		t.Fatal(err)
 	}
 }
+
+func deploymentDefaultTolerations() []corev1.Toleration {
```
iam-veeramalla, line 511: 👍

test/e2e/gitopsservice_test.go
```diff
@@ -845,6 +847,57 @@ var _ = Describe("GitOpsServiceController", func() {
 		})
 	})
 
+	Context("Verify Configuring Infrastructure NodeSelector ", func() {
```
iam-veeramalla, line 850: Thanks for adding the tests :)

common/common.go
saumeya: Sure Abhishek, will remove this comment

api/v1alpha1/gitopsservice_types.go
saumeya: I agree

iam-veeramalla:
Thanks for the PR @saumeya !!
/LGTM
/approve

iam-veeramalla (approved):

PR #227, opened 2021-09-28
chore: Increase the default timeout for e2e tests

iam-veeramalla:
**What type of PR is this?**
> /kind bug

The default timeout for `go test` is 10 mins which may lead to timeout and result in test suite failure. 
```
-timeout d
          If a test binary runs longer than duration d, panic.
          If d is 0, the timeout is disabled.
          The default is 10 minutes (10m).
```

NA

Fixes #?
#228 

NA


wtam2018 (commented):
/lgtm
/approve

PR #230, opened 2021-09-28
fix: Add permissions to operator to delete ResourceQuota object on namespace

iam-veeramalla:
**What type of PR is this?**
> /kind bug

This PR ports the changes #223 into master branch.

**Test acceptance criteria**:
NA

Upgrade the operator from v1.2 to v1.3 and see that the resourcequota on openshift-gitops namespace is not deleted.
After this change it should be deleted.


chetan-rns (approved):
/lgtm
/approve

PR #259, opened 2021-12-24
doc: readme fix for quay username placeholder

saumeya:
**What type of PR is this?**
> /kind documentation

Current readme uses `redhat-developer` for creating and pushing docker image. As most users don't have access to redhat-developer quay account, it is better to add placeholder so that they can run the commands using their quay account.


iam-veeramalla (approved):
LGTM

PR #291, opened 2022-03-23
feat: pull argoproj-labs/argocd-operator#555 (RH-SSO enhancements)

iam-veeramalla:
**What type of PR is this?**
> /kind enhancement

pulls https://github.com/argoproj-labs/argocd-operator/pull/555

**This PR adds the below enhancements.**
-------------------------------------------
* Upgrades the default version of RH-SSO to 7.5.1
* Adds support to login with `kube:admin` OpenShift User.
* With this PR, RH-SSO can fetch the group details of OpenShift Users(You should see the group details of the logged in user in the Argo CD console). This will help admins define Group Level RBAC for their OpenShift groups. 
   https://github.com/keycloak/keycloak/pull/8381
* Kube:admin by default you only have. Not working previously. not an OS user.
   https://github.com/keycloak/keycloak/pull/8428
* RHSSO works in proxy enabled cluster with no additional or manual configuration.
   https://github.com/keycloak/keycloak/pull/8559
   
More technical details
------------------------
* Removed hardcoded OIDC client secrets and replaced them by generated random strings. These client secrets do not need to be stored. 
* Added `KeycloakIdentityProviderMapper` to workaround https://github.com/keycloak/keycloak-operator/issues/471. Without the workaround, tokens can't be translated to credentials that contain the group identity.
* Propagated proxy environment variables from operator to Keycloak containers so that outbound calls can go through cluster proxy if cluste proxy is configured.
* Delete OAuth client using foreground propagation policy to ensure that the garbage collector removes all instantiated objects before the TemplateInstance itself disappears.
* Add extra Delete OAuth client calls to workaround https://github.com/openshift/client-go/issues/209
* Add retries to handle RHSSO image upgrade

For more information please refer 
https://github.com/argoproj-labs/argocd-operator/pull/555

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/redhat-developer/gitops-operator
 go 1.16
 
 require (
-	github.com/argoproj-labs/argocd-operator v0.0.16-0.20220120135155-711b6cf9403e
+	github.com/argoproj-labs/argocd-operator v0.0.16-0.20220323075615-6ebfe8431073
 	github.com/coreos/prometheus-operator v0.40.0
 	github.com/go-logr/logr v1.2.0
 	github.com/google/go-cmp v0.5.6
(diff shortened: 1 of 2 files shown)


chetan-rns (approved):
/lgtm
/approve

PR #308, opened 2022-05-12
feat: add default resource exclusions in argo for pipeline run

saumeya:
**What type of PR is this?**

> /kind bug
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

Fixes #?

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test

### DIFF
--- a/controllers/argocd/openshift/openshift.go
+++ b/controllers/argocd/openshift/openshift.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/yaml"
 )
 
 var log = logf.Log.WithName("openshift_controller_argocd")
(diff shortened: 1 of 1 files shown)


controllers/argocd/openshift/openshift.go
```diff
@@ -15,17 +15,26 @@ import (
...
+	APIGroups []string `json:"apiGroups"`
+	Kinds     []string `json:"kinds"`
+	Clusters  []string `json:"clusters"`
+}
 
 func ReconcilerHook(cr *argoprojv1alpha1.ArgoCD, v interface{}, hint string) error {
 
 	logv := log.WithValues("ArgoCD Namespace", cr.Namespace, "ArgoCD Name", cr.Name)
+
+	cr.Spec.ResourceExclusions = getResourceExclusionsforPipeline()
```
jopit, line 36: Won't this overwrite whatever the user has specified for ResourceExclusions?

Besides, I don't think the operator should be modifying the Argo CD CR.  I think the right way to fix this is to update the argocd-operator's `reconcileArgoConfigMap` (https://github.com/argoproj-labs/argocd-operator/blob/8ef20934780a758f24188ffedb15053beed4f72e/controllers/argocd/configmap.go#L320) function to add a call to `applyReconcilerHook()`, passing in the configmap, and then in the hook code in the gitops-operator, add the resource exclusions to the configmap, being careful to preserve any additional ones set by the user.

jopit (changes requested):

controllers/argocd/openshift/openshift.go
saumeya: Thanks for the review John, I'll make these changes

saumeya:
Creating a PR to address this via a different approach in the ClusterServiceVersion yaml. Updating the configmap without the CR would make this default setting hidden from the user.

PR #321, opened 2022-05-25
Update csv for resource exclusion

saumeya:
**What type of PR is this?**
/kind enhancement

ArgoCD prunes TaskRun and PipelineRun objects if pipeline is manged by it. To avoid the pruning, an exclusion rule needs to be added explicitly. An example below. The exclusion should happen by default.
```
resourceExclusions: |
    - apiGroups:
      - tekton.dev
      clusters:
      - '*'
      kinds:
      - TaskRun
      - PipelineRun 
```   

https://issues.redhat.com/browse/GITOPS-1876

Fixes #?
GITOPS-1876
**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test

https://user-images.githubusercontent.com/17771352/171176341-69c8a5f2-e687-44de-b2d1-e531be10d2ef.mp4

Create the catalog source image and test by creating a new instance from the console.
NOTE : This will only work for Argo instance created from console not CLI

### DIFF
--- a/bundle/tests/scorecard/config.yaml
+++ b/bundle/tests/scorecard/config.yaml
@@ -12,6 +12,9 @@ stages:
     labels:
       suite: basic
       test: basic-check-spec-test
+    storage:
+      spec:
+        mountPath: {}
   - entrypoint:
     - scorecard-test
     - olm-bundle-validation
@@ -19,6 +22,9 @@ stages:
     labels:
       suite: olm
       test: olm-bundle-validation-test
+    storage:
+      spec:
+        mountPath: {}
   - entrypoint:
     - scorecard-test
     - olm-crds-have-validation
@@ -26,6 +32,9 @@ stages:
     labels:
       suite: olm
       test: olm-crds-have-validation-test
+    storage:
+      spec:
+        mountPath: {}
   - entrypoint:
     - scorecard-test
     - olm-crds-have-resources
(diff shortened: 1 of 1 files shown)


bundle/tests/scorecard/config.yaml
```diff
@@ -12,59 +12,38 @@ stages:
     labels:
       suite: basic
       test: basic-check-spec-test
-    storage:
```
iam-veeramalla, line 15: @saumeya I think this is not intentional ? What is the version of Operator-SDK that you are using ?

saumeya: thanks, yes I am using an old one `operator-sdk version: "v1.10.1"`

bundle/manifests/gitops-operator.clusterserviceversion.yaml
```diff
@@ -102,6 +102,7 @@ metadata:
                 }
               }
             },
+            "resourceExclusions": "- apiGroups:\n  - tekton.dev\n  clusters:\n  - '*'\n  kinds:\n  - TaskRun\n  - PipelineRun        \n",
```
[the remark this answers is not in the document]
saumeya: Yes, I have tested it and this works!

chetan-rns:
The `resourcesExclusions` will only be set when the CR is created from the Operator console and doesn't have any effect on CRs created declaratively. Are we okay with restricting it to the console?

iam-veeramalla:
Thanks @saumeya , This PR will be automatically merged once the tests get passed

saumeya:
> The `resourcesExclusions` will only be set when the CR is created from the Operator console and doesn't have any effect on CRs created declaratively. Are we okay with restricting it to the console?

Yes, I confirmed with @wtam2018 about the exception

PR #358, opened 2022-08-03
feat: add custom node selector to GitopsServiceCR 

saumeya:
**What type of PR is this?**
 /kind enhancement

This PR allows the user to add custom nodeSelectors in their GitopsService CR, which will allow them to add selctors to their default instance workloads.

It is also dependent on the merge of https://github.com/argoproj-labs/argocd-operator/pull/747 which adds the default Linux nodeSelector to argocd operator workloads

Fixes GITOPS-2164

**Test acceptance criteria**:

* [x] Unit Test
* [x] E2E Test
Kuttl Tests- https://gitlab.cee.redhat.com/gitops/operator-e2e/-/merge_requests/87


api/v1alpha1/gitopsservice_types.go
```diff
@@ -30,6 +30,8 @@ type GitopsServiceSpec struct {
 	RunOnInfra bool `json:"runOnInfra,omitempty"`
 	// Tolerations allow the pods to schedule onto nodes with matching taints
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+	// NodeSelector is a field of PodSpec, it is a map of key value pairs used for node selection
```
jannfis, line 33: I think this should give a little more clear explanation, like the one for `runOnInfra`. I.e. state that this applies to the default workloads of GitOps operator.

This will be shown to users upon `oc explain`, so indirectly is user documentation :)

controllers/gitopsservice_controller.go
```diff
@@ -330,6 +331,9 @@ func (r *ReconcileGitopsService) reconcileDefaultArgoCDInstance(instance *pipeli
 			NodeSelector: common.InfraNodeSelector(),
 		}
 	}
+	if len(instance.Spec.NodeSelector) > 0 {
```
jannfis, line 334: I was wondering, how does this behave in combination with `runOnInfra` set to true, or with `tolerations` set to any value.

To me it looks like they overwrite each other. For example, if `toleations` and `nodeSelector` is set, `nodeSelector` will be overwritten by `tolerations` below.

Is this expected?

jannfis (commented):
I have some questions, please see below.

Also, I think we'd want a kuttl test for this.

controllers/gitopsservice_controller.go
saumeya: These are actually working on the nodePlacement property in the argocd CR, so there is a check to see if that property exists in the CR already, if it does then it just appends or else creates. So they won't overwrite each other when multiple options are set.

saumeya:
Kuttl Tests- https://gitlab.cee.redhat.com/gitops/operator-e2e/-/merge_requests/87

jopit (approved):
LGTM

PR #377, opened 2022-09-30
Support console api being optional

jopit:
Signed-off-by: John Pitman <jpitman@redhat.com>

Support console api being optional

**What type of PR is this?**

> /kind bug
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

Fixes #?

**Test acceptance criteria**:

* [x] Unit Test
* [ ] E2E Test

### DIFF
--- a/controllers/argocd_controller.go
+++ b/controllers/argocd_controller.go
@@ -30,6 +30,7 @@ import (
 	console "github.com/openshift/api/console/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	"github.com/redhat-developer/gitops-operator/common"
+	"github.com/redhat-developer/gitops-operator/controllers/util"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -86,6 +87,10 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	if err := util.InspectCluster(); err != nil {
+		setupLog.Info("unable to inspect cluster")
+	}
```
jannfis, line 92: Can we safely assume that the availability of the console API will never change during the lifecycle of the cluster?

For example, can the console API be provisioned to the cluster, after it has been installed without the API being available? If this case, the Operator would need to be restarted to pick up the API's availability.

jopit: Not sure what to do here.  We could check for the existence of the console API every time we reconciled, but that would still require a reconcile to be triggered before the operator would pick up the API's availability.  It would also introduce a bit of a performance slowdown for everyone (but maybe that would be negligible).  I don't know of any way to get kubernetes to alert us when new API is added to the system.

@jannfis do you have any suggestions?

jannfis: I think it's fine to keep as is, knowing about this possible caveat. If ConsoleLink API is something that can added/removed at will (and not only at installation time), I think it's a fair workaround to have users simply restart the Operator.

controllers/argocd_controller.go
```diff
@@ -117,6 +118,11 @@ func (r *ReconcileArgoCDRoute) Reconcile(ctx context.Context, request reconcile.
 	reqLogger := logs.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling ArgoCD Route")
 
+	if !util.IsConsoleAPIFound() {
+		reqLogger.Info("Skip argocd route reconcile: OpenShift Console API not found")
+		return reconcile.Result{}, nil
+	}
+
```
jannfis, line 125: I think the route should be reconciled regardless of the availability of ConsoleLink API?

jopit: This reconcile method doesn't reconcile the route, it just updates the consolelink with the existing route, or deletes the console link if the route doesn't exist.

jannfis: Ah. Now as you say it, and looking further down the code... :) 

Then maybe only the log message (and also the existing ones!) are misleading. Because this method doesn't reconcile a route, but a ConsoleLink.

PR #387, opened 2022-10-18
Gitops 1323 config dynamic plugin

ciiay:
**What type of PR is this?**
/kind enhancement

GITOPS-1323, pull gitops-plugin image into operator.

Fixes #? [GITOPS-1323](https://issues.redhat.com/browse/GITOPS-1323)

**Test acceptance criteria**:

* [x] Unit Test
* [ ] E2E Test

1. Create a docker image base off this branch code and install it to a cluster
2. Install pipelines-operator
3. Apply argocd manifests
4. Observe the Dynamic Plugin is available on Developer/Environment - Dynamic


bundle/manifests/pipelines.openshift.io_gitopsservices.yaml
```diff
@@ -34,13 +34,19 @@ spec:
           spec:
             description: GitopsServiceSpec defines the desired state of GitopsService
             properties:
```
jopit, line 36: Not sure why this NodeSelector code is part of this PR?

controllers/gitopsservice_controller.go
```diff
@@ -344,6 +349,15 @@ func (r *ReconcileGitopsService) reconcileDefaultArgoCDInstance(instance *pipeli
 			defaultArgoCDInstance.Spec.NodePlacement.NodeSelector = argocdutil.AppendStringMap(defaultArgoCDInstance.Spec.NodePlacement.NodeSelector, instance.Spec.NodeSelector)
 		}
 	}
+	if len(instance.Spec.NodeSelector) > 0 {
```
jopit, line 352: Again, should this NodeSelector related code be part of this PR?

bundle/manifests/pipelines.openshift.io_gitopsservices.yaml
ciiay: Thanks for the review. I got some code wrong when I rebased my branch. It should be fixed now.

controllers/gitopsservice_controller.go
ciiay: Thanks for the catch. This should be fixed now.

jopit (changes requested):

controllers/consoleplugin.go
```diff
@@ -0,0 +1,405 @@
...
+	kubeAppLabelInstance         = "app.kubernetes.io/instance"
+	kubeAppLabelPartOf           = "app.kubernetes.io/part-of"
+	kubeAppLabelRuntimeNamespace = "app.kubernetes.io/runtime-namespace"
+	httpdConfigMapName           = "httpd-cfg"
+)
+
+func pluginDeployment() *appsv1.Deployment {
+	consolePluginImage := os.Getenv(pluginImageEnv)
+	if consolePluginImage == "" {
+		consolePluginImage = "quay.io/redhat-developer/gitops-console-plugin"
```
iam-veeramalla, line 46: This should be a variable or a constant and split into two parts

Image := quay.io/redhat-developer/gitops-console-plugin 
version := "<whatever is the version>"

controllers/consoleplugin.go
```diff
@@ -0,0 +1,405 @@
...
+					},
+				},
+				Resources: corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceMemory: resourcev1.MustParse("128Mi"),
+						corev1.ResourceCPU:    resourcev1.MustParse("250m"),
+					},
+					Limits: corev1.ResourceList{
+						corev1.ResourceMemory: resourcev1.MustParse("256Mi"),
+						corev1.ResourceCPU:    resourcev1.MustParse("500m"),
```
iam-veeramalla, line 88: How did we derive the resource requests and limits values ? Any research ?

controllers/consoleplugin.go
```diff
@@ -0,0 +1,405 @@
...
+	kubeAppLabelRuntimeNamespace = "app.kubernetes.io/runtime-namespace"
+	httpdConfigMapName           = "httpd-cfg"
+)
+
+func pluginDeployment() *appsv1.Deployment {
+	consolePluginImage := os.Getenv(pluginImageEnv)
+	if consolePluginImage == "" {
+		consolePluginImage = "quay.io/redhat-developer/gitops-console-plugin"
+	}
+	podSpec := corev1.PodSpec{
```
iam-veeramalla, line 48: Can you please test this pod specification/deployment on a 4.12 cluster which is based on k8s v1.25 to verify if we need to include podSecurity ?

https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

controllers/consoleplugin.go
```diff
@@ -0,0 +1,405 @@
...
+	template := corev1.PodTemplateSpec{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels: map[string]string{
+				kubeAppLabelApp: gitopsPluginName,
+			},
+		},
+		Spec: podSpec,
+	}
+	var replicas int32 = 1
+	return &appsv1.Deployment{
```
iam-veeramalla, line 128: Can this be split into different functions ?

podSpec := getPluginPodSpec()
getPluginDeployment(podSpec)

iam-veeramalla (changes requested):

controllers/consoleplugin.go
ciiay: I copied these settings from kam.go. How should I decide the value?

iam-veeramalla: You can create a  resource quota of 20 CPU and 20 GB on openshift-gitops namespace
after that, all the pods in the namespace can be created only if thy have resource requests and limits in their specification, just try to keep the resource limits as low as posible and run trail and error

this just for your understanding.

As of this PR, the requests and limits seems to be low already. I am ok with them

jaideepr97:
@ciiay you will have to have all your commits signed off on to pass the DCO check

controllers/consoleplugin.go
ciiay: @jaideepr97 In what case we need to put default value in `common.go` ? Or just all default value should be in `common.go`?

jaideepr97: Yeah, I think all hard coded default values should be tracked in common.go or defaults.go

That way it is easier to make changes in the future and we don't have to hunt down where individual strings are stored

iam-veeramalla (approved):
/lgtm
/approve

PR #398, opened 2022-11-21
gitops-2415 added reconcile plugin condition

ciiay:
**What type of PR is this?**

/kind enhancement

To avoid multiple "Environments" button showing up in left menu on Developer tab in DevConsole.

Fixes GITOPS-2415

**Test acceptance criteria**:

* [x] Unit Test
* [ ] E2E Test

1. On OCP 4.15 or greater, observe that Console Plugin section is showing up on GitOps-operator installation page. Only one "Environments" button exists on Developer tab left menu.
2. On OCP 4.14 or lower, observe that Console Plugin section is not showing up on GitOps-operator installation page. Only one "Environments" button exists on Developer tab left menu.


jaideepr97:
Hi @ciiay @keithchong 
Are we still going with the approach of making the dynamic plugin optionally enabled (but disabled by default) under 4.15?
In this PR it seems like the plugin will not work for clusters under 4.15 at all

ciiay:
> Hi @ciiay @keithchong Are we still going with the approach of making the dynamic plugin optionally enabled (but disabled by default) under 4.15? In this PR it seems like the plugin will not work for clusters under 4.15 at all

Hi Jaideep, thanks for the review. Yes, going forward we will only reconcile dynamic plugin resources when OCP is 4.15 or higher version. On OCP 4.14 or lower, user will still be able to use static plugin. Please refer to the [GITOPS-2415](https://issues.redhat.com/browse/GITOPS-2415) and Keith's comment in [GITOPS-2369](https://issues.redhat.com/browse/GITOPS-2369)

jaideepr97:
@ciiay according to that comment we will disable it by default, but users should have the option to enable it - and we should document that enabling the dynamic plugin means there will be 2 environment tabs rendered in the console

This will require changes in the CRD to add a new field to enable/disable teh dynamic plugin, and the operator would need to be able to reconcile the plugin based on that setting in the CR

ciiay:
> @ciiay according to that comment we will disable it by default, but users should have the option to enable it - and we should document that enabling the dynamic plugin means there will be 2 environment tabs rendered in the console
> 
> This will require changes in the CRD to add a new field to enable/disable teh dynamic plugin, and the operator would need to be able to reconcile the plugin based on that setting in the CR

For now the SDK is not supporting to set Console Plugin default to Enable, so we have to only reconcile it once the static plugin gets removed on OCP 4.15.

Where else to document the 2 environment buttons rendered in the console besides GITOPS-2369 and GITOPS-2415?

jaideepr97:
> > @ciiay according to that comment we will disable it by default, but users should have the option to enable it - and we should document that enabling the dynamic plugin means there will be 2 environment tabs rendered in the console
> 
> This will require changes in the CRD to add a new field to enable/disable teh dynamic plugin, and the operator would need to be able to reconcile the plugin based on that setting in the CR
> 
> For now the SDK is not supporting to set Console Plugin default to Enable, so we have to only reconcile it once the static plugin gets removed on OCP 4.15.
> 
> Where else to document the 2 environment buttons rendered in the console besides GITOPS-2369 and GITOPS-2415?

I'm not talking about enabling by default...I'm saying last I knew we were going to keep it disabled by default for < 4.15 but were still going to support enabling the dynamic plugin via the CR on OCP < 4.15 if the user explicitly wished to do so
I'm just wondering if that has changed

ciiay:
> > > @ciiay according to that comment we will disable it by default, but users should have the option to enable it - and we should document that enabling the dynamic plugin means there will be 2 environment tabs rendered in the console
> > > This will require changes in the CRD to add a new field to enable/disable teh dynamic plugin, and the operator would need to be able to reconcile the plugin based on that setting in the CR
> > 
> > 
> > For now the SDK is not supporting to set Console Plugin default to Enable, so we have to only reconcile it once the static plugin gets removed on OCP 4.15.
> > Where else to document the 2 environment buttons rendered in the console besides GITOPS-2369 and GITOPS-2415?
> 
> I'm not talking about enabling by default...I'm saying last I knew we were going to keep it disabled by default for < 4.15 but were still going to support enabling the dynamic plugin via the CR on OCP < 4.15 if the user explicitly wished to do so I'm just wondering if that has changed

There are two definition of Enable.
1. Creating resources for dynamic plugin -> this has to happen only when static plugin gets removed which is planned on OCP 4.15
2. Enable in Console Plugin section on gitops-operator installation page -> this will be available as long as dynamic plugin resources are created

We want to keep only one “Environments” button, so to avoid two buttons showing up, we only create dynamic plugins when static plugin is no longer there

controllers/gitopsservice_controller_test.go
```diff
@@ -303,7 +334,7 @@ func TestReconcile_consoleAPINotFound(t *testing.T) {
 	s := scheme.Scheme
 	addKnownTypesToScheme(s)
 
-	fakeClient := fake.NewFakeClient(newGitopsService())
+	fakeClient := fake.NewFakeClient(util.NewClusterVersion("4.15.1"), newGitopsService())
```
jaideepr97, line 337: @ciiay is this required here?

controllers/gitopsservice_controller_test.go
```diff
@@ -500,7 +531,7 @@ func TestReconcile_InfrastructureNode(t *testing.T) {
 			Tolerations: deploymentDefaultTolerations(),
 		},
 	}
-	fakeClient := fake.NewFakeClient(gitopsService)
+	fakeClient := fake.NewFakeClient(util.NewClusterVersion("4.12.1"), gitopsService)
```
jaideepr97, line 534: is this required here?

controllers/gitopsservice_controller_test.go
ciiay: @jaideepr97 Yes, it's needed, because I’m checking the OCP version in Reconcile function, if I don’t set it in the unit tests, it will through error and unit tests will fail. Or is there a better way to solve the problem?

controllers/gitopsservice_controller_test.go
ciiay: Same as the previous one, if I don't set it here the unit test fails.

iam-veeramalla:
@ciiay @jaideepr97 @keithchong 
Is there a way to use the new dynamic plugin on OCP 4.12 ? If yes, how ?

jaideepr97:
@iam-veeramalla that is what i have asked above, it seems the decision is to only allow creation of dynamic plugin resources starting 4.15  (not even provide the option to enable it by the user)

so im actually not sure how anyone would be able to test this until we have a 4.15 cluster 

cc @ciiay @keithchong

controllers/gitopsservice_controller.go
```diff
@@ -224,7 +225,21 @@ func (r *ReconcileGitopsService) Reconcile(ctx context.Context, request reconcil
...
+		return reconcile.Result{}, err
+	}
+
+	v1, err := version.NewVersion(OCPVersion)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	v2, err := version.NewVersion("4.15.0")
+	if v1.LessThan(v2) {
+		return reconcile.Result{}, err
```
jaideepr97, line 239: @ciiay I don't think we need to return an error here, we can just have 

```
if !v1.LessThan(v2) {
     r.reconcilePlugin(instnace, request)
    }
```

controllers/gitopsservice_controller_test.go
jaideepr97: please see https://github.com/redhat-developer/gitops-operator/pull/398/files#r1029254556

controllers/gitopsservice_controller_test.go
jaideepr97: https://github.com/redhat-developer/gitops-operator/pull/398/files#r1029254556

ciiay:
New PR created https://github.com/redhat-developer/gitops-operator/pull/402. Closing this PR due to rebase and sign off issue.

PR #401, opened 2022-11-22
Skip test based on condition

varshab1210:
Signed-off-by: varshab1210 <vab@redhat.com>

**What type of PR is this?**

/kind failing-test

Skipping constantly failing test on 4.9 based on condition 

Fixes #?

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test


scripts/openshiftci-presubmit-all-tests.sh
```diff
@@ -24,5 +24,8 @@ cp $KUBECONFIG $TMP_DIR/kubeconfig
 chmod 640 $TMP_DIR/kubeconfig
 export KUBECONFIG=$TMP_DIR/kubeconfig
 
+export OCP_VERSION="$(oc version)"
+echo $OCP_VERSION
+
```
wtam2018, line 29: If we can avoid the dependency of the env var, that would be better.  Folks may run these tests from an IDE without calling the script.

varshab1210: Thanks @wtam2018, that's a good point. But **os.Getenv("OCP_VERSION")** will return an empty string when **OCP_VERSION** is not set (if someone is running the test locally using make target) and we have seen the test passes on the local environment even on 4.9 cluster so I don't think that should be a problem. WDYT?

PR #415, opened 2023-01-05
Making Kuttl tests compatible with Openshift CI

varshab1210:
**What type of PR is this?**

> /kind enhancement

Currently we have tests bound to ginkgo framework as part of Openshift CI which has less test coverage and is more error prone. This PR aims at making Kuttl tests capable of running as part of CI along with ginkgo tests (until we migrate all ginkgo tests to Kuttl). This PR also includes certain new files that will help us create operator from scratch

Fixes #?

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test


scripts/openshift-CI-kuttl-tests.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# Copy kubeconfig to temporary kubeconfig file and grant read and Write permission to temporary kubeconfig file
+TMP_DIR=$(mktemp -d)
+cp $KUBECONFIG $TMP_DIR/kubeconfig
+chmod 640 $TMP_DIR/kubeconfig
+export KUBECONFIG=$TMP_DIR/kubeconfig
+cp $KUBECONFIG /go/src/github.com/redhat-developer/gitops-operator/kubeconfig
+
+# Ensuring proper installation
+pod=gitops-operator-controller-manager && oc get pods `oc get pods --all-namespaces | grep $pod | head -1 | awk '{print $2}'` -n openshift-operators -o yaml || true
```
rjeczkow, line 21: this line would return `0` even if the `$pod` is not found (`grep $pod`).
```
$ true && false || true
$ echo $?
0
```

I think we should just remove `|| true`

scripts/openshift-CI-kuttl-tests.sh
```diff
@@ -0,0 +1,41 @@
...
+  namespace: test-argocd
+EOF
+
+sleep 60s
+
+oc get pods -n test-argocd
+
+# oc wait --for=condition=Ready -n test-argocd pod --timeout=15m  -l 'app.kubernetes.io/name in (argocd-application-controller,argocd-redis,argocd-repo-server,argocd-server)' 
+
+echo ">> Running tests on ${TARGET}"
```
rjeczkow, line 41: where the `${TARGET}` variable comes from ?

openshift-ci/build-root/Dockerfile
```diff
@@ -3,3 +3,30 @@
 FROM registry.ci.openshift.org/openshift/release:golang-1.18
 
 RUN yum -y install dnf httpd-tools
```
rjeczkow, line 5: Seems this line is not needed any more

PR #427, opened 2023-02-01
feat: Add default channel and bundle channels for 1.8

iam-veeramalla:
**What type of PR is this?**
> /kind enhancement

I realized that the below annotations are missing on the operator.  These annotations are useful for building the operator similar to the Red Hat downstream operator. 

I found out that they were missing during the render_templates implementation. With render_templates we will implement the manifests and metadata folders of the bundle directly from this repo.
```
  operators.operatorframework.io.bundle.channel.default.v1: latest
  operators.operatorframework.io.bundle.channels.v1: latest,gitops-1.8
```

This PR just adds annotations that are required by the downstream build.


jaideepr97:
/lgtm
Thanks @iam-veeramalla

jaideepr97 (approved):

PR #435, opened 2023-02-03
Add target and manifests for Core operator installation

bluengo:
**What type of PR is this?**

/kind enhancement
> /kind bug
> /kind cleanup
> /kind failing-test
> /kind documentation
> /kind code-refactoring

You can find the new changes in the documentation [here](https://github.com/bluengo/gitops-operator/blob/GITOPSRVCE-376/docs/OpenShift%20GitOps%20Usage%20Guide.md?plain=1#L78-L86)

More information in issue [GITOPSRVCE-376](https://issues.redhat.com/browse/GITOPSRVCE-376) 

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test

Pick up a successful GitOps operator nightly build, and run the new make target to install the core operator in your current OpenShift cluster:

`make deploy-core IMG=<gitops_operator_bundle_image>` 

Remember that this method, as the old `make deploy`, assumes that the cluster already has credentials to access the registry where the images within the bundle are stored.

### DIFF
--- a/config/core/default_env_patch.yaml
+++ b/config/core/default_env_patch.yaml
@@ -1,0 +1,18 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+      - name: manager
+        env:
+        - name: DISABLE_DEFAULT_ARGOCD_INSTANCE
+          value: "true"
+        - name: DISABLE_DEFAULT_ARGOCD_CONSOLELINK
+          value: "true"
+        - name: DISABLE_DEX
+          value: "true"
+  
--- a/Makefile
+++ b/Makefile
@@ -151,6 +151,10 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/default | kubectl delete --ignore-not-found=true -f -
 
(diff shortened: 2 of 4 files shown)


config/core/default_env_patch.yaml
```diff
@@ -0,0 +1,18 @@
...
+  template:
+    spec:
+      containers:
+      - name: manager
+        env:
+        - name: DISABLE_DEFAULT_ARGOCD_INSTANCE
+          value: "true"
+        - name: DISABLE_DEFAULT_ARGOCD_CONSOLELINK
+          value: "true"
+        - name: DISABLE_DEX
```
ishitasequeira, line 16: I think we need to disable KAM deployments too.

ishitasequeira (changes requested):

config/core/default_env_patch.yaml
bluengo: Hi @ishitasequeira! According to the [documentation](https://github.com/redhat-developer/gitops-operator/blob/master/docs/OpenShift%20GitOps%20Usage%20Guide.md#setting-environment-variables), there is no way to disable KAM by using the container Environment.
If the Core operator cannot provide this mechanism, I'm afraid that it would need to be either amended or shipped along with KAM components.

ishitasequeira: As discussed over chat, let's create a separate story for `DISABLE_KAM` flag and get this PR merged.

ishitasequeira:
The changes look good to me 
/lgtm
Once the CI is fixed, we could merge this

ishitasequeira (approved):

PR #446, opened 2023-02-22
fix: Fix monitoring test

jaideepr97:
**What type of PR is this?**

 /kind bug
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

Fixes #?

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test


test/openshift/e2e/sequential/1-077_validate_workload_status_monitoring_alert/02-verify-alert.yaml
```diff
@@ -1,13 +1,13 @@
-# apiVersion: kuttl.dev/v1beta1
-# kind: TestStep
-# commands:
-# - script: sleep 10m
-# - script: |
-#     jq '.data.alerts' <<< "$(oc -n openshift-monitoring exec -c prometheus prometheus-k8s-0 -- curl -s   'http://localhost:9090/api/v1/alerts')" > alerts.json
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+- script: sleep 10m
```
varshab1210, line 4: **sleep 10m** would wait for 10m even when the condition is met and delay other test executions.Can we have a polling mechanism to avoid this?

test/openshift/e2e/sequential/1-077_validate_workload_status_monitoring_alert/01-install.yaml
```diff
@@ -1,12 +1,12 @@
...
-# spec:
-#   applicationSet: 
-#     image: test-image
-#   monitoring:
-#     enabled: true
+apiVersion: argoproj.io/v1alpha1
+kind: ArgoCD
+metadata:
+  name: openshift-gitops
+  namespace: openshift-gitops
```
varshab1210, line 5: As @rjeczkow mentioned the test leaves applicationset-controller in ImagePullBackOff (to check if the alert is firing). We should probably not use openshift-gitops namespace for it?

rjeczkow: If using `openshift-gitops` namespace is not necessary (The alerts would fire for any other argo-cd instance I believe) we could use argo-cd in other namespace and place this test in 'parallel' suite because it wouldn't modify the default argo-cd instance any more. Then we could keep the `sleep 10` since it wouldn't block other tests execution.

test/openshift/e2e/sequential/1-077_validate_workload_status_monitoring_alert/01-assert.yaml
```diff
@@ -1,39 +1,39 @@
...
+          labels:
+            severity: critical
+        - alert: RepoServerNotReady
+          for: 1m
+          labels:
+            severity: critical
+        - alert: ApplicationSetControllerNotReady
+          for: 5m
+          labels:
+            severity: warning
```
rjeczkow, line 27: I'm thinking if this shouldn't be critical actually.

test/openshift/e2e/sequential/1-077_validate_workload_status_monitoring_alert/02-verify-alert.yaml
```diff
@@ -1,13 +1,13 @@
...
+- script: |
+    jq '.data.alerts' <<< "$(oc -n openshift-monitoring exec -c prometheus prometheus-k8s-0 -- curl -s   'http://localhost:9090/api/v1/alerts')" > alerts.json
 
-#     result=$(jq --arg deployment "openshift-gitops-applicationset-controller" --arg namespace $NAMESPACE '.[] | select(.labels.alertname == "ApplicationSetControllerNotReady" and .state == "firing")' alerts.json)
+    result=$(jq --arg deployment "openshift-gitops-applicationset-controller" --arg namespace "openshift-gitops" '.[] | select(.labels.alertname == "ApplicationSetControllerNotReady" and .state == "firing")' alerts.json)
```
rjeczkow, line 8: Why do we need `--arg deployment` and `--arg namespace` here? It would define `deployment` and `namespace` variables for jq but no filtering would happen.  I'm wondering if they should rather go to `select` as `.labels.deployment` and `.labels.namespace`

varshab1210:
Closing this, please reopen if required

PR #479, opened 2023-03-29
fix: add argoproj.io rules in rbac for non-olm install

saumeya:
**What type of PR is this?**
 /kind bug

Issue - https://github.com/redhat-developer/gitops-operator/issues/148

Fixes #?

https://github.com/redhat-developer/gitops-operator/issues/148

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test


controllers/argocd/openshift/openshift.go
```diff
@@ -92,12 +92,37 @@ func ReconcilerHook(cr *argoprojv1alpha1.ArgoCD, v interface{}, hint string) err
...
+				policyRules = append(policyRules, getPolicyRuleForNonOlmArgoCDApplicationController()...)
+			}
 			o.Rules = policyRules
 		}
 	}
 	return nil
 }
 
+// checks if argoproj.io rules exist in application controller role, if not return false
+func argoExists(policies []rbacv1.PolicyRule) (result bool) {
```
drpaneas, line 105: please make the name more descriptive. This might also mean the ArgoCD instance exists.

controllers/argocd/openshift/openshift.go
```diff
@@ -92,12 +92,37 @@ func ReconcilerHook(cr *argoprojv1alpha1.ArgoCD, v interface{}, hint string) err
...
+		if contains(policy.APIGroups, "argoproj.io") {
+			result = true
+			break
+		}
+	}
+	return result
+}
+
+// utility function to check for string in a slice
+func contains(s []string, e string) bool {
```
drpaneas, line 117: Pick better names. Having code that reads like ` a == e` is not ideal. For example use `s` for string and `slice` for the slice. Or something like this. Same thing for the name of the function, `contains` is very generic. Some ideas: `stringExistsInSlice`

drpaneas (changes requested):
Hey @saumeya  thanks for the PR. I have reviewed it and just a few minor requests:
1. The naming conventions you are using are very generic, please try to make them more specific tailored to the scope/goal of the function.
2. Please provide unit-tests as well.

Many thanks :D

varshab1210:
/retest

CI platform issue

varshab1210:
/test v4.8-kuttl-parallel

platform issue

controllers/argocd/openshift/openshift.go
```diff
@@ -92,12 +92,37 @@ func ReconcilerHook(cr *argoprojv1alpha1.ArgoCD, v interface{}, hint string) err
...
 		}
 	}
 	return nil
 }
 
+// checks if argoproj.io rules exist in application controller role, if not return false
+func argoprojRulesExists(policies []rbacv1.PolicyRule) (result bool) {
+	result = false
+	for _, policy := range policies {
+		if stringExistsInSlice(policy.APIGroups, "argoproj.io") {
```
jaideepr97, line 108: @saumeya I don't think it is sufficient to only check for the API group here. There could be a situation where the API group is present but permissions were only added for some resources, not others, or some of the verbs are missing

I think it would be better to check whether the exact ruleset is present or not

saumeya: I think these permissions are added by OLM when the operator is installed - so all the rules will exist not just a few verbs/resources etc. What else is the process of adding policy rules?

jaideepr97: This PR is explicitly dealing with the case where OLM is not present, so we cannot rely on that 
If there is a situation where someone creates a custom role where they include the argoproj group but dont add all the necessary resource/verb permissions correctly, this check will not be able to catch it

jaideepr97 (commented):
@saumeya could we have another test that invokes the hook and tests that the role will be appropriately modified to contain argoproj permissions?

controllers/argocd/openshift/openshift.go
saumeya: Please check the logic, if it looks good or needs further improvement.

saumeya:
> @saumeya could we have another test that invokes the hook and tests that the role will be appropriately modified to contain argoproj permissions?

working on it, I will have to refactor the way the k8sclient is invoked within the hook, any suggestions how to improve that?

jaideepr97:
>> @saumeya could we have another test that invokes the hook and tests that the role will be appropriately modified to contain argoproj permissions?
>
> working on it, I will have to refactor the way the k8sclient is invoked within the hook, any suggestions how to improve that?

Yeah actually, now that I look at it more closely, this code is bad practice. We should not be initializing a new k8s client there. We probably can't even fake a client to test this case 

The only way this would work that I can think of would also involve changes in argocd-operator in terms how to gain accesss to a client

controllers/argocd/openshift/openshift.go
jaideepr97: I don't think this is the best approach to do this 
I would suggest something like this instead:

```
func main() {

	a := PolicyRule{
		APIGroups: []string{"grp1", "grp2"},
		Resources: []string{"r1", "r2"},
		Verbs:     []string{"v1"},
	}

	b := PolicyRule{
		APIGroups: []string{"grp2", "grp1"},
		Resources: []string{"r1", "r2"},
		Verbs:     []string{"v1"},
	}

	sort.Strings(b.APIGroups) 
        sort.Strings(b.Resources)
        sort.Strings(b.Verbs)

	fmt.Println(reflect.DeepEqual(a, b))
        // returns true 

}

```

jaideepr97: I don't think this is the best approach to do this 
I would suggest something like this instead:

```
import (
     ...
     "sort"
     "reflect"
     ...
)

...

func argoprojPermissionsExist (policies []rbacv1.PolicyRule) (result bool) {

        permissionsExist := false
        
	for _, policy := range policies {
		sort.Strings(policy.APIGroups)
		sort.Strings(policy.Resources)
		sort.Strings(policy.Verbs)
		
		if(reflect.DeepEqual(policy, getArgoProjPolicyRule()) {
		    permissionsExist = true
		    break
		}
		
	}
	return permissionsExist
}
```

saumeya:
> > > @saumeya could we have another test that invokes the hook and tests that the role will be appropriately modified to contain argoproj permissions?
> > 
> > 
> > working on it, I will have to refactor the way the k8sclient is invoked within the hook, any suggestions how to improve that?
> 
> Yeah actually, now that I look at it more closely, this code is bad practice. We should not be initializing a new k8s client there. We probably can't even fake a client to test this case
> 
> The only way this would work that I can think of would also involve changes in argocd-operator in terms how to gain accesss to a client. That would not be appropriate within this PR

Exactly, that is why I have created a test for the function that checks the rules, because adding a client would change the way the hook gets registered. I agree that the code must be improved. 
I'll create a ticket for it in the product backlog but for time being the normal test seems the best alternate.

controllers/argocd/openshift/openshift.go
saumeya: It is just that when I look at the policies added in the role by the operator, the resources are split across rules and in that case the target will not be equal to the policy in the check.

jaideepr97: I'm not sure I understand the concern
As far as I can tell the only times that resources from the same API group are split across rules is when we explicitly want to set different permissions for different resources (eg retrieve pods but not secrets from "")

I don't see that being the case with argoproj.io resources

saumeya: when i real the role in our operator - it reads like this
`- verbs:
      - '*'
    apiGroups:
      - argoproj.io
    resources:
      - applications
  - verbs:
      - '*'
    apiGroups:
      - argoproj.io
    resources:
      - applicationsets
  - verbs:
      - '*'
    apiGroups:
      - argoproj.io
    resources:
      - appprojects
  - verbs:
      - '*'
    apiGroups:
      - argoproj.io
    resources:
      - argocds
  - verbs:
      - create
      - update
      - patch
      - delete
    apiGroups:
      - argoproj.io
    resources:
      - applications
  - verbs:
      - create
      - update
      - patch
      - delete
    apiGroups:
      - argoproj.io
    resources:
      - applicationsets
  - verbs:
      - create
      - update
      - patch
      - delete
    apiGroups:
      - argoproj.io
    resources:
      - appprojects
  - verbs:
      - create
      - update
      - patch
      - delete
    apiGroups:
      - argoproj.io
    resources:
      - argocds`

saumeya:
Just appended the argoproj rules to the existing policy.

We could check the existing permissions (see code snippet) for these resources, but it would be an overkill, adding these policies again doesn't conflict in olm installation and hence would be a simpler efficient solution

`// create the AuthorizationV1 clientset
	authClient := clientset.AuthorizationV1()

	ssar := &authorizationv1.SelfSubjectAccessReview{
		Spec: authorizationv1.SelfSubjectAccessReviewSpec{
			ResourceAttributes: &authorizationv1.ResourceAttributes{
				Namespace: namespace,
				Verb:      "*",
				Group:     "argoproj.io",
				Resource:  "applications",
			},
		},
	}

	// send the SelfSubjectAccessReview request to the Kubernetes API
	ssarResponse, err := authClient.SelfSubjectAccessReviews().Create(context.TODO(), ssar, metav1.CreateOptions{})
	if err != nil {
		panic(err.Error())
	}

	// check the response to see if the user is allowed to perform the requested action
	if ssarResponse.Status.Allowed {
		fmt.Println("User has permission to access the resource")
	} else {
		fmt.Println("User does not have permission to access the resource")
	}`

ishitasequeira (approved):
The changes look good!! The role gets updated with added permissions.

jaideepr97:
> The changes look good!! The role gets updated with added permissions.

@ishitasequeira did you test this through a non-olm installation?

saumeya:
> > The changes look good!! The role gets updated with added permissions.
> 
> @ishitasequeira did you test this through a non-olm installation?

Verified that it works and adds the rules with the non-olm script

jaideepr97 (approved):
lgtm then, thanks @saumeya !

iam-veeramalla:
Hi @varshab1210 , is this failure due to the CI ?

iam-veeramalla:
Thanks @saumeya , Ignoring the 4.8 tests as they are out of support.

iam-veeramalla (approved):

PR #531, opened 2023-06-09
Support migration to non-OLM installation and using regctl for extracting CSV info from bundle image

anandf:
**What type of PR is this?**
/kind enhancement

Improve the manifest based installation script (non-OLM installation) by removing the dependency on either `docker` or `podman` and instead using a lightweight binary `regctl` to extract the `ClusterServiceVersion` from within the bundle image. Since this binary gets automatically installed and uninstalled if not present on the host system, its easy for users to run the installation script without going through the overheads of setting up `docker` or `podman` on their host systems.

Fixes #?

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test

### DIFF
--- a/hack/non-olm-install/README.md
+++ b/hack/non-olm-install/README.md
@@ -7,19 +7,25 @@
 The `install-gitops-operator.sh` script supports two methods of installation.
 1. Using operator and component images set as environment variables (default method)
 2. Derive the operator and component images from the `ClusterServiceVersion` manifest present in the operator bundle
-**Note**: This method requires podman or docker binary to be available in the PATH environment variable. Use environment variables `USE_BUNDLE_IMG`, `BUNDLE_IMG` and `DOCKER` for this method of installation
+**Note**: Use environment variables `USE_BUNDLE_IMG`, `BUNDLE_IMG` for this method of installation
 
 
 ### Known issues and work arounds
 
-1. Missing RBAC access to update CRs in `argoproj.io` domain 
+1. Missing RBAC access to update CRs in `argoproj.io` domain
 
-Issue: (Now fixed)
+Affected versions:
+- 1.7.4 and older versions
+- 1.8.3 and older versions
 
+Fixed versions:
+- 1.8.4 and later versions
+- 1.9.0 and later versions
+
+Issue:
 https://github.com/redhat-developer/gitops-operator/issues/148
 
 Workaround:
-
 Run the following script to create the required `ClusterRole` and `ClusterRoleBinding`
 
 ```
@@ -32,7 +38,7 @@ ${KUBECTL} apply -f https://raw.githubusercontent.com/redhat-developer/gitops-op
 **Note**: If the above binaries are not present, the script installs them to temporary work directory and are removed once the script execution is complete.
 - bash (v5.0 or later)
 - git (v2.39.1 or later)
-- podman (v4.4.4 or later) or docker (Note: Required only if operator and component images need to be derived from a bundle image)
+- wget (v1.21.3 or later)
 
 ### Environment Variables
 The following environment variables can be set to configure various options for the installation/uninstallation process.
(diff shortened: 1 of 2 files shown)


hack/non-olm-install/README.md
```diff
@@ -7,19 +7,25 @@
 The `install-gitops-operator.sh` script supports two methods of installation.
 1. Using operator and component images set as environment variables (default method)
 2. Derive the operator and component images from the `ClusterServiceVersion` manifest present in the operator bundle
```
saumeya, line 9: In which use case would a user need to install the operator via the operator-bundle?

hack/non-olm-install/install-gitops-operator.sh
```diff
@@ -394,6 +403,66 @@ function print_info() {
...
+  if [ "${element}" == "null" ]; then
+    echo "[INFO] No custom config present in Subscription"
+    return
+  fi
+
+  # for each custom env, convert it to key=value combination.
+  while IFS=$'\t' read -r name value _; do
+    echo "Setting $name=$value"
+    echo "export $name=$value" >> ${WORK_DIR}/migrate_env.sh
+  done < <(yq e '.[] | [.name, .value] | @tsv' ${WORK_DIR}/env_overrides.yaml)
```
saumeya, line 461: Where does the `env_overrides.yaml` file get used

saumeya (commented):
Ran the script and migration was successful.

ishitasequeira (changes requested):

hack/non-olm-install/install-gitops-operator.sh
```diff
@@ -394,6 +403,66 @@ function print_info() {
...
+  remove_subscription
+  remove_installed_csv
+  wait_for_olm_removal
+  if [ -f ${WORK_DIR}/migrate_env.sh ];then
+    echo "Sourcing env variables used for customizing subscription"
+    source ${WORK_DIR}/migrate_env.sh
+  fi
+  apply_kustomize_manifests
+  # Check pod status if it becomes ready
+  check_pod_status_ready gitops-operator-controller-manager
```
ishitasequeira, line 418: As discussed offline, we do not really need rollback.. `check_pod_status_ready` calls `rollback` if the pod failed to come up..

hack/non-olm-install/README.md
anandf: We can use `BUNDLE_IMG` to extract the component build URLs from inside the `ClusterServiceVersion` manifest inside the bundle.

hack/non-olm-install/install-gitops-operator.sh
anandf: Implemented the rollback feature for migration. If the migration fails, the script switches back to the OLM installation, so that the error can be corrected and re-run.

hack/non-olm-install/install-gitops-operator.sh
anandf: This file will be used during the migration. Certain configuration settings are set in `Subscription`, like `DISABLE_DEFAULT_ARGOCD_INSTANCE` which needs to be carried over to the non OLM installation as well. The `env-overrides.yaml` file contains the env configuration set in `Subscription`

hack/non-olm-install/README.md
```diff
@@ -7,19 +7,25 @@
...
 
 
 ### Known issues and work arounds
 
-1. Missing RBAC access to update CRs in `argoproj.io` domain 
+1. Missing RBAC access to update CRs in `argoproj.io` domain
 
-Issue: (Now fixed)
+Affected versions:
+- 1.7.4 and older versions
```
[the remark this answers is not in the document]
anandf: Added  `1.7.x` and `1.8.x` releases which are actively maintained, so that the fix can be done on a later bug fix versions, like the fix can be made available in `v1.8.4` and `v1.7.5`

hack/non-olm-install/README.md
```diff
@@ -180,5 +187,11 @@ rm -f ${oldauth} ${newauth}
...
-OPERATOR_REGISTRY=brew.registry.redhat.io OPERATOR_REGISTRY_ORG=rh-osbs GITOPS_OPERATOR_VER=v99.9.0-88 IMAGE_PREFIX="openshift-gitops-1-" ./install-gitops-operator.sh -i
-```
\ No newline at end of file
+OPERATOR_REGISTRY=brew.registry.redhat.io OPERATOR_REGISTRY_ORG=rh-osbs GITOPS_OPERATOR_VER=v99.9.0-<build_number> IMAGE_PREFIX="openshift-gitops-1-" ./install-gitops-operator.sh -i
+```
+
+#### Migrate from an OLM based install to non-OLM based installation (nightly-build)
+
+```
+OPERATOR_REGISTRY=brew.registry.redhat.io OPERATOR_REGISTRY_ORG=rh-osbs IMAGE_PREFIX=openshift-gitops-1- GITOPS_OPERATOR_VER=v99.9.0-<build_number> ./install-gitops-operator.sh -m
```
ishitasequeira, line 196: Should we add an example of Migrate with basic default params as well?

ishitasequeira (approved):
LGTM!! Tested the Install, Migrate and Uninstall workflows successfully

PR #540, opened 2023-06-15
[WIP] test: Rewrite 1-018_validate_disable_default_instance kuttl test

svghadi:
**What type of PR is this?**
/kind failing-test

Rewrite 1-018_validate_disable_default_instance kuttl test to prevent flaky behaviour.

Fixes [GITOPS-2829](https://issues.redhat.com/browse/GITOPS-2829)

### DIFF
--- a/test/openshift/e2e/ignore-tests/sequential/1-018_validate_disable_default_instance/02-assert.yaml
+++ b/test/openshift/e2e/ignore-tests/sequential/1-018_validate_disable_default_instance/02-assert.yaml
@@ -1,9 +1,0 @@
-apiVersion: apps/v1
-kind: Deployment
-metadata:
-  name: gitops-operator-controller-manager
-  namespace: openshift-operators
-status:
-  readyReplicas: 1
-  replicas: 1
-  availableReplicas: 1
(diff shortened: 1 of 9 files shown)


svghadi:
@varshab1210 - Can you review it? 
I tested the PR thrice, didn't encounter failure due to `1-018_validate_disable_default_instance` test.

test/openshift/e2e/sequential/1-018_validate_disable_default_instance/01-patch-subscription-disable-false.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: kuttl.dev/v1beta1
```
varshab1210, line 1: Hi @svghadi, Why do we need to set the variable false initially?

svghadi: A test to verify that argocd resources are not deleted if `DISABLE_DEFAULT_ARGOCD_INSTANCE` is set to non `true` value. I didn't want to add extra step to verify this hence added it in initial step. We can remove it if you think it's not required.

varshab1210: In my opinion we should add this step after setting it to true and before removing the env variable. Because, I have seen in the past, setting it to false and then true triggers the reconciliation (flaky nature)

svghadi: Done

varshab1210:
/lgtm
Thanks @svghadi

PR #553, opened 2023-07-07
updated Argo CD 2.7.6 CRDs and argocd-operator

reginapizza:
**What type of PR is this?**
 /kind enhancement

Argocd-operator was [updated with the latest CRDs](https://github.com/argoproj-labs/argocd-operator/pull/944) from [Argo CD 2.7.6](https://github.com/argoproj/argo-cd/tree/v2.7.6/manifests/crds), this PR pulls in those changes and updates CRDs to match. 

NA

I know that the changes to CRDs include adding in typos but that's what's in upstream... :upside_down_face:

### DIFF
--- a/bundle/manifests/argoproj.io_applications.yaml
+++ b/bundle/manifests/argoproj.io_applications.yaml
@@ -1428,7 +1428,7 @@ spec:
                           as part of automated sync (default: false)'
                         type: boolean
                       selfHeal:
-                        description: 'SelfHeal specifies whether to revert resources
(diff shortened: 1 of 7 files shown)


jaideepr97:
@reginapizza the version still says 1.8.0 in the CSV 
we should probably have updated that earlier

iam-veeramalla:
@jaideepr97 the version update in the CSV is handled by render templates :)

reginapizza:
Is it ok to merge this PR? It keeps failing because of what I believe are just flakey tests

PR #580, opened 2023-08-01
update 1-031 to check for parallel component version fixes:2641

anandrkskd:
**What type of PR is this?**

/kind failing-test

This PR updates and fixes ignored test 1-031 with updated version of argocd, helm and dex. 

Fixes # GITOPS-2641

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test

- ~any cluster with  openshfit-gitops version installed~
- run 
```
kubectl kuttl test ./test/openshift/e2e/parallel   --config ./test/openshift/e2e/parallel/kuttl-test.yaml --test  1-031_validate_toolchain
```


test/openshift/e2e/parallel/1-031_validate_toolchain/01-check.yaml
```diff
@@ -0,0 +1,106 @@
...
+      )
+
+      redisVersion=$(
+        oc -n openshift-gitops exec $redis_pod -- \
+          redis-server -v | awk -F '=' '{ print $2 }' | cut -d'.' -f 1
+      )
+
+      if test "${kustomizeVersion}" != "${expected_kustomizeVersion}"; then
+        echo "Kustomize version mismatch. Should be ${expected_kustomizeVersion}, is ${kustomizeVersion}"
+        exit 1
```
varshab1210, line 81: Hi @anandrkskd 
Lets  catch **exit 1** after setting a  flag for each condition if a mismatch happens

test/openshift/e2e/parallel/1-031_validate_toolchain/01-check.yaml
```diff
@@ -0,0 +1,106 @@
...
+  - script: |
+      set -e
+      set -o pipefail
+
+      # These variables need to be maintained according to the component matrix: https://spaces.redhat.com/display/GITOPS/GitOps+Component+Matrix
+      expected_kustomizeVersion='v5.0.1'
+      expected_helmVersion='v3.11.2'
+      expected_argocdVersion='v2.7.6'
+      expected_dexVersion='v2.35.1'
+      expected_redisVersion='6'
```
varshab1210, line 13: Hi @anandrkskd 
Could you also please add [this](https://gitlab.cee.redhat.com/gitops/operator-e2e/-/commit/3b9ac9fb6f1583edd5b9ef300cf2e40f0e347f89) change?
Thanks

anandrkskd:
As per discussion with Siddhesh, while building gitops-operator container for tests we are fetching dex Version and Redis version from argocd [here](https://github.com/argoproj-labs/argocd-operator/blob/master/common/defaults.go#L232), To make it close to release build we need to update component versions to the one we are using from midstream repos.

This also posses a problem for future, i.e. it will get out of sync and fail, once we update the version used in midstream.
WDYT? @iam-veeramalla, @varshab1210

svghadi:
IMO enabling this test is a good way to ensure that upstream and downstream versions are in sync. Whenever this test fails we know there is a mis-match somewhere in versions and we can update accordingly and won't go un-noticed.

anandrkskd:
>IMO enabling this test is a good way to ensure that upstream and downstream versions are in sync. Whenever this test fails we know there is a mis-match somewhere in versions and we can update accordingly and won't go un-noticed.

But, shouldn't we start from making changes from argocd-operator repo instead of mid-strem repo? or do we have some reason due to which this is not possible?

svghadi:
For any versions related changes, usually it is upstream first so `argocd`->`argocd-operator`->`gitops-operator`. However I am not sure in this case, why downstream has different dex version. Maybe @varshab1210 or @iam-veeramalla knows the reason. 

That being said version difference makes sense if we rely on base image in downstream for the package. For example, we install git via base image repos so most of the times it will differ from upstream as the base image is different. However in case of dex, we build the container so we are in full control of the version.

varshab1210:
LGTM

PR #586, opened 2023-08-17
Fixed the default ArgoCD role for default instance openshift-gitops

anandf:
**What type of PR is this?**
/kind enhancement
> /kind bug
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

This PR is to make the default ArgoCD instance with name `openshift-gitops` in namespace `openshift-gitops` to have restricted default permissions. With default `readonly` permission, any non-admin user will be able to view the Application and other manifest resource managed by ArgoCD which can contain sensitive information. By setting the default permission to `''` this read only access can be avoided.

Fixes #?

**Test acceptance criteria**:

* [x] Unit Test
* [ ] E2E Test

### DIFF
--- a/controllers/argocd/argocd.go
+++ b/controllers/argocd/argocd.go
@@ -28,6 +28,9 @@ import (
 var (
 	defaultAdminPolicy = "g, system:cluster-admins, role:admin\ng, cluster-admins, role:admin\n"
 	defaultScope       = "[groups]"
+
+	//The policy.default property in the argocd-rbac-cm ConfigMap.
+	defaultArgoCDRole = ""
 )
 
 // resource exclusions for the ArgoCD CR.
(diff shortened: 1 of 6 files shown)


jaideepr97:
@anandf could this be a breaking change for users that haven't specifically defined their argocd rbac because the default policy was already letting them do what they needed ? Not that that is a reason not to make this change

anandf:
> @anandf could this be a breaking change for users that haven't specifically defined their argocd rbac because the default policy was already letting them do what they needed ? Not that that is a reason not to make this change

I agree that this change can cause surprise in behaviour especially if customers were leveraging this default read only role. Would it be sufficient to mention this change in behaviour in the release notes ? Please not that this change being applicable only for the default gitops instance and other customer managed `ArgoCD` instances will continue working as before.

jaideepr97 (dismissed):
LGTM, thanks @anandf

PR #589, opened 2023-08-23
Add OpenShift Dashboards for GitOps

gnunn1:
**What type of PR is this?**

/kind enhancement

This provisions a set of dashboards in the OpenShift console (Observer > Dashboards) to provide monitoring for GitOps. These dashboards are based on the upstream Argo CD dashboards but have been modified to work with the OpenShift console which has a reduced feature set compared to Grafana.

https://issues.redhat.com/browse/GITOPS-1767

Fixes #?

GITOPS-1767

**Test acceptance criteria**:

* [*] Unit Test
* [ ] E2E Test

Install the GitOps Operator on an OpenShift cluster and then navigate to Observe > Dashboards in the console. In the dropdown list of dashboards note that three new dashboards are now available:

* GitOps Overview
* GitOps Components
* GitOps gRPC


controllers/argocd_metrics_controller.go
```diff
@@ -292,6 +308,79 @@ func (r *ArgoCDMetricsReconciler) createPrometheusRuleIfAbsent(namespace string,
...
+					// Should we return an error or continue with other dashboards?
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func newDashboard(filename string, namespace string) (*corev1.ConfigMap, error) {
```
jaideepr97, line 359: can we call this `newDashboardConfigMap` instead? Makes it easier to follow the code knowing what the function is actually doing without getting into it

controllers/argocd_metrics_controller.go
```diff
@@ -292,6 +308,79 @@ func (r *ArgoCDMetricsReconciler) createPrometheusRuleIfAbsent(namespace string,
 	return err
 }
 
+func (r *ArgoCDMetricsReconciler) createDashboardsIfAbsent(reqLogger logr.Logger) error {
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: dashboardNamespace}, &corev1.Namespace{})
+	if err != nil {
+		reqLogger.Info("OpenShift dashboards are not installed, skipping dashboard installation",
```
jaideepr97, line 314: could we rephrase this log statement? Could be a little confusing to users
maybe it could say something like "Monitoring dashboards are not supported on this cluster, skipping dashboard installation"? 

Lmk what you think

jaideepr97:
Thanks @gnunn1, I tested this out and it works great!
One other point I would mention is maybe we can move the `dashboards` folder out of the `controllers` folder and keep it at the same level as the controllers folder

gnunn1:
Moving dashboards outside of controllers doesn't work for me, there are some limitations around go embed with respect to where the content is located. 
```
The go:embed directive only works with files that are accessible from the current working directory or a subdirectory of it. This means that you cannot use the go:embed directive to embed files outside of your project directory or in a parent directory.
```
https://www.iamyadav.com/blogs/a-guide-to-embedding-static-files-in-go

gnunn1:
I have made the requested changes plus added the ability for it to update the dashboard configmap if the data element doesn't match the expectation.

controllers/argocd_metrics_controller.go
```diff
@@ -292,6 +308,94 @@ func (r *ArgoCDMetricsReconciler) createPrometheusRuleIfAbsent(namespace string,
...
+
+			if errors.IsNotFound(err) {
+				reqLogger.Info("Creating new dashboard",
+					"Namespace", dashboard.Namespace, "Name", dashboard.Name)
+				err := r.Client.Create(context.TODO(), dashboard)
+				if err != nil {
+					reqLogger.Error(err, "Error creating a new dashboard",
+						"Namespace", dashboard.Namespace, "Name", dashboard.Name)
+					// Should we return an error or continue with other dashboards?
+					return err
```
jaideepr97, line 366: I think we should continue with other dashboards here, unless they only make sense in context of each other 
If that is not that case and there is value in each dashboard independently, we should not return error here

same for line 332 as well

controllers/argocd_metrics_controller.go
```diff
@@ -292,6 +308,91 @@ func (r *ArgoCDMetricsReconciler) createPrometheusRuleIfAbsent(namespace string,
...
+	}
+
+	for _, entry := range entries {
+		reqLogger.Info("Processing dashboard", "Namespace", dashboardNamespace, "Name", entry.Name())
+
+		if !entry.IsDir() {
+			dashboard, err := newDashboardConfigMap(entry.Name(), dashboardNamespace)
+			if err != nil {
+				reqLogger.Info("There was an error creating dashboard ", "Namespace", dashboardNamespace, "Name", entry.Name())
+				return err
```
jaideepr97, line 332: @gnunn1 please remove this line as well

jaideepr97 (approved):
/lgtm
/approve

Thanks a lot @gnunn1!

PR #601, opened 2023-09-12
Updated default version to 1.9.2 and tools kustomize, regctl, yq

anandf:
**What type of PR is this?**
/kind bug

Non OLM installation fails as the latest version of CRD manifests creates resources with `creationTimestamp="null"` this causes the builds to fail with error 
```
error: unable to decode "/tmp/gitops-operator-install-dxw9u9c/kustomize-build-output.yaml": parsing time "null" as "2006-01-02T15:04:05Z07:00": cannot parse "null" as "2006"
```
Also, updated the default version and dependent tool versions - yq, kustomize and regctl

Fixes #?

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test

### DIFF
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -12,7 +12,7 @@ resources:
 # Comment the following 4 lines if you want to disable
 # the auth proxy (https://github.com/brancz/kube-rbac-proxy)
 # which protects your /metrics endpoint.
-- auth_proxy_service.yaml
+#- auth_proxy_service.yaml
 - auth_proxy_role.yaml
 - auth_proxy_role_binding.yaml
-- auth_proxy_client_clusterrole.yaml
+#- auth_proxy_client_clusterrole.yaml
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -12,7 +12,7 @@ resources:
 # Comment the following 4 lines if you want to disable
 # the auth proxy (https://github.com/brancz/kube-rbac-proxy)
 # which protects your /metrics endpoint.
-#- auth_proxy_service.yaml
+# - auth_proxy_service.yaml
 - auth_proxy_role.yaml
 - auth_proxy_role_binding.yaml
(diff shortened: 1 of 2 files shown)


hack/non-olm-install/install-gitops-operator.sh
```diff
@@ -177,6 +177,7 @@ resources:
   - https://github.com/redhat-developer/gitops-operator/config/crd?ref=$GIT_REVISION&timeout=90s
   - https://github.com/redhat-developer/gitops-operator/config/rbac?ref=$GIT_REVISION&timeout=90s
   - https://github.com/redhat-developer/gitops-operator/config/manager?ref=$GIT_REVISION&timeout=90s
+  - https://raw.githubusercontent.com/redhat-developer/gitops-operator/master/config/rbac/auth_proxy_service.yaml
```
jaideepr97, line 180: @anandf I would suggest instead of adding this file, we should uncomment the auth_proxy_service in the kustomization file and pick it up from there

Please also add `config/prometheus` to this list so that the artifacts necessary for operator metrics are included in non-olm installation

anandf: Made the suggested changes. Please review.

config/rbac/kustomization.yaml
```diff
@@ -12,7 +12,7 @@ resources:
 # Comment the following 4 lines if you want to disable
 # the auth proxy (https://github.com/brancz/kube-rbac-proxy)
 # which protects your /metrics endpoint.
-# - auth_proxy_service.yaml
+- auth_proxy_service.yaml
 - auth_proxy_role.yaml
 - auth_proxy_role_binding.yaml
-# - auth_proxy_client_clusterrole.yaml
+- auth_proxy_client_clusterrole.yaml
```
jaideepr97, line 18: @anandf we don't need this uncommented, we have replaced this clusterrole with the role defined in `config/prometheus`

anandf: ack. Removed all changes done to this file `config/rbac/kustomization.yaml`

PR #622, opened 2023-11-20
fix: reconcile operator metrics service monitor based on installation namespace 

jaideepr97:
**What type of PR is this?**

 /kind bug
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

This PR adds some logic to reconcile the operator metrics service monitor. It makes sure that the `serverName` field in the endpoint TLS config is updated to reflect the namespace that the operator is installed in rather than always pointing to `openshift-gitops-operator`. This fixes the issue of alerts being generated when the operator is installed in any namespace other than `openshfit-gitops-operator` 

Fixes #?

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test

- build an operator image from this branch
- install operator in a namespace other than `openshift-gitops-operator` such as `openshift-operators`
- let installation complete
- after installation is done, examine the svc monitor called `openshift-gitops-operator-metrics-monitor` in the same namespace
- look at `spec.endpoints[0].tlsConfig.serverName` and verify that it is structured as `openshift-gitops-operator-metrics-service.<installation-namespace>.svc` (like `openshift-gitops-operator-metrics-service.openshift-operators.svc` for example)


controllers/argocd_metrics_controller.go
```diff
@@ -274,6 +282,42 @@ func (r *ArgoCDMetricsReconciler) createServiceMonitorIfAbsent(namespace string,
...
+	operatorNS := string(data)
+	desiredMetricsServerName := operatorMetricsServiceName + "." + operatorNS + ".svc"
+
+	existingServiceMonitor := &monitoringv1.ServiceMonitor{}
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: operatorMetricsMonitorName, Namespace: operatorNS}, existingServiceMonitor)
+
+	if err == nil {
+		changed := false
+
+		if existingServiceMonitor.Spec.Endpoints[0].TLSConfig.ServerName != desiredMetricsServerName {
```
anandf, line 302: Can `.Spec.Endpoints` be `nil` or point to an empty array ?

controllers/argocd_metrics_controller.go
```diff
@@ -274,6 +282,42 @@ func (r *ArgoCDMetricsReconciler) createServiceMonitorIfAbsent(namespace string,
...
+			existingServiceMonitor.Spec.Endpoints[0].TLSConfig.ServerName = desiredMetricsServerName
```
anandf, line 303: If there are no items in the `Endpoints` array, We need to create a new `TLSConfig` and add that element isn't it ? Would it not rather better to override the entire desired Endpoint ?
```existingServiceMonitor.Spec.Endpoints = Endpoint[]{&TLSConfig{ServerName: desiredMetricsServerName}}```
Also can there be a scenario where there could be multiple endpoints ?

File: controllers/argocd_metrics_controller.go (hunk 1 above)
Comment (jaideepr97) [reply controllers/argocd_metrics_controller.go:302]: not unless someone manually deletes the endpoint from the service monitor for some reason
the sevice monitor that we deploy at install time already has an endpoint configured out of the box

File: controllers/argocd_metrics_controller.go (hunk 2 above)
Comment (jaideepr97) [reply controllers/argocd_metrics_controller.go:303]: technically yes, I think. having multiple end points would just mean this service monitor is watching multiple services. In our case that won't happen because this service monitor only watches a single metrics service, so there will only be one endpoint 

This PR's scope was restricted to just the changes needed to address a known issue that we want to fix in 1.11, that's why the change is rather specific to just reconciling that one serverName field 

if you think it would be better to just reconcile the entire servicemonitor I am open to that as well
at the moment I don't expect any other fields to get modified though

File: controllers/argocd_metrics_controller.go [hunk 3]
```diff
@@ -274,6 +282,42 @@ func (r *ArgoCDMetricsReconciler) createServiceMonitorIfAbsent(namespace string,
...
+			changed = true
```
anandf, line 304: Instead of setting this flag and later updating, can we not call `r.Client.Update` here ?

controllers/argocd_metrics_controller.go
```diff
@@ -274,6 +282,42 @@ func (r *ArgoCDMetricsReconciler) createServiceMonitorIfAbsent(namespace string,
...
+		}
+
+		if changed {
+			return r.Client.Update(context.TODO(), existingServiceMonitor)
+		}
+	} else if err != nil {
```
anandf, line 310: This code can simplified with `if..else`. If err, return `err` early. Else, check for desired server name and update it.

anandf (approved):

controllers/argocd_metrics_controller.go
jaideepr97: updated

controllers/argocd_metrics_controller.go
jaideepr97: updated, thanks

controllers/argocd_metrics_controller.go
```diff
@@ -170,6 +173,11 @@ func (r *ArgoCDMetricsReconciler) Reconcile(ctx context.Context, request reconci
 		return reconcile.Result{}, err
 	}
 
+	err = r.reconcileOperatorMetricsServiceMonitor(reqLogger)
+	if err != nil {
+		return reconcile.Result{}, nil
```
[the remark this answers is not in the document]
jaideepr97: updated, thanks

ciiay:
Hi @jaideepr97 , can you take another look at changes in this PR? It seems that
```
go test `go list ./... | grep -v test` -coverprofile cover.out
``` 
command will fall with this update.

PR #644, opened 2024-01-19
Change TLS termination to Reencrypt and Redirect on Insecure

supernovae:
**What type of PR is this?**

> /kind bug

When this operator is installed, the default route created is pass through so users get an invalid certificate warning since the gitops secret for tls is a generic.

Lots of corporate managed browsers do NOT let users bypass invalid certificates through accepting cert or typing "thisisunsafe" 

Default behavior for fresh install should be a working route using the apps* route.

No documentation necessary, without this PR, we should have a documentation bug. So if this isn't approved, we should update our docs on how to either update the certificate or update the CRD to use reencrypt.

Fixes # GITOPS-3839

**Test acceptance criteria**:

* [ ] E2E Test

Install operator, configure gitops CRD and open default link to ArgoCD - should no longer get SSL warning or blocked browser warning.

### DIFF
--- a/controllers/kam.go
+++ b/controllers/kam.go
@@ -144,8 +144,8 @@ func newRouteForCLI() *routev1.Route {
 			TargetPort: intstr.IntOrString{IntVal: portTLS},
 		},
 		TLS: &routev1.TLSConfig{
-			Termination:                   routev1.TLSTerminationPassthrough,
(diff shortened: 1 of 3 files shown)


olivergondza:
@supernovae, please take a look at resolvign the conflict. Thanks

PR #696, opened 2024-05-02
Fix Kuttl E2E tests: Switch Rollouts tests to sequential, change Role to ClusterRole

jgwest:
**What type of PR is this?**

/kind failing-test

Update kuttl tests:
- Argo Rollouts tests cannot run in parallel, as multiple cluster-scoped RolloutManagers on a cluster is an unsupported scenario
- Argo Rollouts default to ClusterRoles, rather than Roles


jgwest:
Merging, but there are still some failing tests: this is likely due to https://github.com/redhat-developer/gitops-operator/pull/694, which has a cyclical dependency on this PR (#696). So I'm merging #696, to unblock #694.

PR #702, opened 2024-05-14
feat: make metrics scraping optional

saumeya:
**What type of PR is this?**
 /kind enhancement

To make metrics scraping optional in GitOps Operator

Fixes #?

**Test acceptance criteria**:

* [x] Unit Test
* [x] E2E Test

Edit the spec.monitoring field to false to disable scraping of metrics, like so

apiVersion: argoproj.io/v1beta1
kind: ArgoCD
metadata:
 name: argocd-sample
 namespace: default
spec:
 monitoring:
   disableMetrics: true

     
OR patch the resource using cli, like so

oc patch argocd argocd-sample -n default --type='json' -p='[{"op": "replace", "path": "/spec/monitoring/disableMetrics", "value": true}]'

To enable metrics for your instance, just change the spec.monitoring field to false, this will create the required role, rolebindings and service monitors and add the cluster-monitoring label to the namespace resource. 
Disabling it will delete these resources and will also remove the cluster-monitoring label from the namespace resource.


controllers/argocd_metrics_controller.go
```diff
@@ -110,72 +110,135 @@ func (r *ArgoCDMetricsReconciler) Reconcile(ctx context.Context, request reconci
...
+			}
+		} else {
+			namespace.Labels[clusterMonitoringLabel] = "true"
+			err = r.Client.Update(ctx, &namespace)
+			if err != nil {
+				reqLogger.Error(err, "Error updating namespace",
+					"Namespace", namespace.Name)
+				return reconcile.Result{}, err
+			}
+		}
```
svghadi, line 134: Is this redundant block intentional?

saumeya: Yes, it the user wants to enable monitoring after disabling it, then in that case we need to re-assign the label and update the namespace. In the first if condition we are creating the label map.

svghadi: Okay. In that case, we can simplify the if  block with something like 

```
        labelVal, exists := namespace.Labels[clusterMonitoringLabel]

	if argocd.Spec.Monitoring.Enabled {
		if !exists || labelVal != "true" {
			if namespace.Labels == nil {
				namespace.Labels = make(map[string]string)
			}
			namespace.Labels[clusterMonitoringLabel] = "true"
			err = r.Client.Update(ctx, &namespace)
			if err != nil {
				reqLogger.Error(err, "Error updating namespace",
					"Namespace", namespace.Name)
			}
		} 
```
WDYT?

saumeya: makes sense!

controllers/argocd_metrics_controller.go
```diff
@@ -109,73 +109,128 @@ func (r *ArgoCDMetricsReconciler) Reconcile(ctx context.Context, request reconci
 	}
 
 	const clusterMonitoringLabel = "openshift.io/cluster-monitoring"
-	_, exists := namespace.Labels[clusterMonitoringLabel]
-	if !exists {
-		if namespace.Labels == nil {
-			namespace.Labels = make(map[string]string)
+	labelVal, exists := namespace.Labels[clusterMonitoringLabel]
+
+	if argocd.Spec.Monitoring.Enabled {
```
svghadi, line 114: Just adding this as a note, no immediate action required.  [argocd.Spec.Monitoring.Enabled](https://github.com/argoproj-labs/argocd-operator/blob/4bd23cbd1a30f71eff97a4941b7485c914032e45/api/v1beta1/argocd_types.go#L687-L694) is used to control  creation of Prometheus alerting rules. With this change, we are reusing this field to enable/disable metric scraping, which makes sense to me. However, the current change won't support use-cases where user wants metric scraping enabled but alerting disabled OOTB. I am not sure how common this use-case is, but if we decide to support it in future, we could introduce a new field `argocd.Spec.Monitoring.Alerts` to control alerts rule creation.

svghadi (changes requested):
Thanks @saumeya. I tested out the change. Just a small suggestion, rest all looks good.

saumeya:
Thanks for reviewing @svghadi

controllers/argocd_metrics_controller.go
```diff
@@ -282,6 +337,18 @@ func (r *ArgoCDMetricsReconciler) createServiceMonitorIfAbsent(namespace string,
...
 
+func (r *ArgoCDMetricsReconciler) deleteServiceMonitor(name string, namespace string, reqLogger logr.Logger) error {
+
+	err := r.Client.Delete(context.TODO(), &monitoringv1.ServiceMonitor{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: name}})
+	if err != nil {
+		reqLogger.Error(err, "Error deleting servicemonitor",
+			"Namespace", namespace)
+		return err
+	}
+	return nil
```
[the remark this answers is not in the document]
svghadi: Hi @saumeya, I just realized that return on not found error would lead to incomplete cleanup due to how the code is structured. Can we update the logic to below in all the places? This will ensure cleanup is proper if first resource is deleted successfully but subsequent resources are deleted in the next reconciliation cycle. 

```go
if !apierrors.IsNotFound(err) {
  reqLogger.Error(err, "Error deleting servicemonitor", "Namespace", namespace)
  return err
}
```

controllers/argocd_metrics_controller.go
```diff
@@ -109,73 +109,135 @@ func (r *ArgoCDMetricsReconciler) Reconcile(ctx context.Context, request reconci
...
+
+			// Delete alert rule
+			err = r.Client.Delete(context.TODO(), &monitoringv1.PrometheusRule{ObjectMeta: metav1.ObjectMeta{Namespace: request.Namespace, Name: alertRuleName}})
+			if err != nil {
+				if !errors.IsNotFound(err) {
+					reqLogger.Error(err, "Error deleting prometheus in ",
+						"Namespace", request.Namespace)
+					return reconcile.Result{}, err
+				}
+			}
```
svghadi, line 238: Are the prometheus rules deleted here the same as created by `.spec.monitoring.enabled` ?
https://argocd-operator.readthedocs.io/en/latest/usage/monitoring/

saumeya: No, this is different

svghadi (dismissed):
/lgtm

PR #707, opened 2024-05-22
fix: secure redis communication[CVE-2024-31989] 

iam-veeramalla:
**What type of PR is this?**
> /kind bug

Fix CVE-2024-31989

It has been discovered that an unprivileged pod in a different namespace on the same cluster could connect to the Redis server on port 6379. Despite having installed the latest version of the VPC CNI plugin on the EKS cluster, it requires manual enablement through configuration to enforce network policies. This raises concerns that many clients might unknowingly have open access to their Redis servers. This vulnerability could lead to Privilege Escalation to the level of cluster controller, or to information leakage, affecting anyone who does not have strict access controls on their Redis instance. This issue has been patched in version(s) 2.8.19, 2.9.15 and 2.10.10.

NA

Fixes CVE-2024-31989


go.mod
```diff
@@ -1,48 +1,50 @@
 module github.com/redhat-developer/gitops-operator
 
-go 1.21
+go 1.22.0
+
+toolchain go1.22.2
```
svghadi, line 5: Was this intentional?

svghadi: Fixed in latest commit

PR #715, opened 2024-05-23
fix: sentinal liveliness issue

iam-veeramalla:
**What type of PR is this?**
> /kind bug

```
Readiness probe failed: AUTH failed: ERR AUTH <password> called without any password configured for the default user. Are you sure your configuration is correct?
```


svghadi (commented):
/lgtm

PR #721, opened 2024-05-30
chore: Update argocd version and manifests

svghadi:

Update argocd version to 2.11 and manifests


config/rbac/role.yaml
```diff
@@ -441,18 +439,6 @@ rules:
   - patch
   - update
   - watch
-- apiGroups:
```
anandf, line 444: These policies are required for Redis related network policies right ?

anandf: Thanks for resolving this.

PR #736, opened 2024-06-12
'run-rollouts-e2e-tests.sh' should fail outright when an individual test fails

jgwest:
**What type of PR is this?**
/kind bug

- At present, the 'run-rollouts-e2e-tests.sh' script should fail when one of the child test suite fails
- The 'set -e' check was missing which should have been present initially, I have added it


chetan-rns (approved):
Thanks @jgwest 

/lgtm

jgwest:
Thanks @chetan-rns!

PR #740, opened 2024-07-03
Fix bug where ArgoCD removes nodePlacement stanza from configuration

Rizwana777:
**What type of PR is this?**
/kind bug

JIRA - https://issues.redhat.com/browse/GITOPS-4185

We were not able to patch the NodePlacement directly through ArgoCD CR, if we do,  it removes the nodePlacement from ArgoCD CR while reconciling because of this https://github.com/redhat-developer/gitops-operator/blob/master/controllers/gitopsservice_controller.go#L475 here `defaultArgoCDInstance.Spec.NodePlacement` is nil as it not getting updated anywhere else in code, `defaultArgoCDInstance.Spec.NodePlacement` is only getting updated if we update it through  GitOpsService CR or else it  update `existingArgoCD.Spec.NodePlacement` as nil because of the reason I have mentioned above  , but if a user wants to add custom nodePlacement he/she should patch nodePlacement through GitOpsService CR and it is working as expected  I see no bug in code and  code is valid https://github.com/redhat-developer/gitops-operator/blob/01251d23020555e3a28d4a69e30718e9d5f0eb4d/controllers/gitopsservice_controller.go#L392

I have updated the code here - https://github.com/redhat-developer/gitops-operator/blob/01251d23020555e3a28d4a69e30718e9d5f0eb4d/controllers/gitopsservice_controller.go#L475
which is checking for nil , the logic mentioned in the PR states that , it will check for `defaultArgoCDInstance.Spec.NodePlacement` not equal to nil (it means nodePlacePlacement is updated through GitOps service CR and the values got updated [here](https://github.com/redhat-developer/gitops-operator/blob/01251d23020555e3a28d4a69e30718e9d5f0eb4d/controllers/gitopsservice_controller.go#L392) in if condition) else it will not enter the if condition to check if `defaultArgoCDInstance.Spec.NodePlacement` is nil (it means  nodePlacePlacement is updated through ArgoCD by user and it should not replace the existingArgoCD with nil value, that is the changes made by user will not get reverted back as mentioned in the bug when it reconciles)

This is what I have came up with for the bug mentioned above, please do let me know your suggestions, thank you

### DIFF
--- a/controllers/gitopsservice_controller.go
+++ b/controllers/gitopsservice_controller.go
@@ -472,9 +472,12 @@ func (r *ReconcileGitopsService) reconcileDefaultArgoCDInstance(instance *pipeli
 			changed = true
 		}
 
-		if !reflect.DeepEqual(existingArgoCD.Spec.NodePlacement, defaultArgoCDInstance.Spec.NodePlacement) {
-			existingArgoCD.Spec.NodePlacement = defaultArgoCDInstance.Spec.NodePlacement
-			changed = true
+		// if user is patching nodePlacement through GitopsService CR, then existingArgoCD NodePlacement is updated.
+		if defaultArgoCDInstance.Spec.NodePlacement != nil {
+			if !reflect.DeepEqual(existingArgoCD.Spec.NodePlacement, defaultArgoCDInstance.Spec.NodePlacement) {
+				existingArgoCD.Spec.NodePlacement = defaultArgoCDInstance.Spec.NodePlacement
+				changed = true
+			}
 		}
 
 		if changed {
--- a/controllers/gitopsservice_controller.go
+++ b/controllers/gitopsservice_controller.go
@@ -423,7 +423,6 @@ func (r *ReconcileGitopsService) reconcileDefaultArgoCDInstance(instance *pipeli
 		}
 	} else {
 		changed := false
-
 		if existingArgoCD.Spec.ApplicationSet != nil {
 			if existingArgoCD.Spec.ApplicationSet.Resources == nil {
 				existingArgoCD.Spec.ApplicationSet.Resources = defaultArgoCDInstance.Spec.ApplicationSet.Resources
@@ -478,6 +477,10 @@ func (r *ReconcileGitopsService) reconcileDefaultArgoCDInstance(instance *pipeli
 				existingArgoCD.Spec.NodePlacement = defaultArgoCDInstance.Spec.NodePlacement
 				changed = true
 			}
+			// Handle the case where NodePlacement should be removed
+		} else if existingArgoCD.Spec.NodePlacement != nil {
+			existingArgoCD.Spec.NodePlacement = defaultArgoCDInstance.Spec.NodePlacement
+			changed = true
 		}
 
 		if changed {
--- a/controllers/argocd_metrics_controller_test.go
+++ b/controllers/argocd_metrics_controller_test.go
@@ -29,7 +29,6 @@ import (
 	is "gotest.tools/assert/cmp"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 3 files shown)


controllers/gitopsservice_controller.go
```diff
@@ -472,9 +472,15 @@ func (r *ReconcileGitopsService) reconcileDefaultArgoCDInstance(instance *pipeli
 			changed = true
 		}
 
-		if !reflect.DeepEqual(existingArgoCD.Spec.NodePlacement, defaultArgoCDInstance.Spec.NodePlacement) {
-			existingArgoCD.Spec.NodePlacement = defaultArgoCDInstance.Spec.NodePlacement
-			changed = true
+		// Checking if default nodePlacement is nil to verify whether the patch is done using GitopService CR or ArgoCD CR
+		// if user is patching nodePlacement through GitopsService CR, then the above written logic for GitopsService will add custom values to defaultArgoCDInstance's nodePlacement,
```
saumeya, line 476: The logic is that unless any change is made in GitopService CR, don't check the nodePlacement spec right?
That will allow users to manage the default instance via ArgoCD CR.

The comment is confusing and not sure if needed.

saumeya:
Also, did you verify why the users are unable to patch the resource via the GitopService CR? why does that not add the required nodeSelectors, maybe there is another bug in that.

controllers/gitopsservice_controller.go
Rizwana777: In the bug, user is trying to patch nodePlacement using `oc patch argocd ...` and logic present is checking for nodePlacement when user tries to patch using gitopsservice `oc patch gitopsservice ...` . Basically in comment it is mentioned that if the user is trying to patch argocd using gitopsservice then `defaultArgoCDInstance.Spec.NodePlacement` will not be nil and it will update argocd(i.e, existingArgoCD.Spec.NodePlacement ) based on that, otherwise `defaultArgoCDInstance.Spec.NodePlacement` will be nil and it will not update the `existingArgoCD.Spec.NodePlacement ` now as I have added If condition here to check this, I have added the explanation in the description, please let me know if its not making sense

Rizwana777:
> Also, did you verify why the users are unable to patch the resource via the GitopService CR? why does that not add the required nodeSelectors, maybe there is another bug in that.

Yes I tried reproducing both the bugs , I reproduced one of the bug where user is trying to patch using argocd (oc patch argocd)  and the second one((oc patch gitopsservice) I did not see this bug, I tried every steps to reproduce it, but not seeing this bug and I am able to patch argocd via GitopsService CR

controllers/gitopsservice_controller.go
anandrkskd: IIUC, what Saumeya is trying to convey is, instead of this comment i.e. explaining how this part of code works, if we can just reduce the comment to what it does. 
Maybe we can just write 

```
     // if user is patching nodePlacement through GitopsService CR, then existingAgoCd NodePlacement is updated.
```

Rizwana777: got it, I will update this, thanks Saumeya and Anand

jgwest (changes requested):
Fix LGTM, could we get some unit tests for this case?

jgwest:
LGTM, rebasing to latest to see if that fixes the test failures

jgwest:
CI issue?

/retest

varshab1210:
Hi @jgwest, there are four sequential test failures that are recurring for this PR so I don't think re trigger would help. Retesting without proper investigation leads to more resource consumption and at the moment we are trying to minimise our cloud cost. I have suggested @Rizwana777 to take a look at the tests that are failing and run them individually with her change to debug further, maybe some tests need updates?

1-071_validate_node_selectors
1-028_validate_run_on_infra
1-020_validate_redis_ha_nonha  (might not be related but worth checking since it fails for this PR only)
1-035_validate_argocd_secret_repopulate (might not be related but worth checking since it fails for this PR only)

Sequential test history: 
https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-redhat-developer-gitops-operator-master-v4.12-kuttl-sequential
https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-redhat-developer-gitops-operator-master-v4.13-kuttl-sequential
https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-redhat-developer-gitops-operator-master-v4.14-kuttl-sequential

jgwest (approved):
LGTM, thanks @Rizwana777, and thanks everyone for your reviews/input!

PR #749, opened 2024-07-18
fix: CVE 2022 41717

Mangaal:
**What type of PR is this?**

 /kind bug

fix  CVE 2022 41717
upgrade dependency golang.org/x/net


go.mod
```diff
@@ -96,6 +96,7 @@ replace (
 	github.com/onsi/gomega => github.com/onsi/gomega v1.14.0
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20220929131120-2bde012f248a
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200325131901-f7baeb993edb
+	golang.org/x/net => golang.org/x/net v0.4.0
```
jgwest, line 99: Isn't this an old version of /x/net?

Mangaal: Hi @jgwest , I realised there was a typo in the version.

Thank you for catching this and correcting it

go.mod
```diff
@@ -96,6 +96,7 @@ replace (
 	github.com/onsi/gomega => github.com/onsi/gomega v1.14.0
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20220929131120-2bde012f248a
 	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20200325131901-f7baeb993edb
+	golang.org/x/net => golang.org/x/net v0.24.0
```
jgwest, line 99: Is this the correct place to put it e.g. within `replace` block? I'm not sure what Go best practice is.  As an example, here are  dependabot commits that was raised that placed/updated it at the top of the file: 
- https://github.com/argoproj-labs/argocd-operator/pull/1347
- https://github.com/argoproj-labs/argocd-operator/pull/1037

And sometimes it is just updated in indirect requires:
- https://github.com/argoproj-labs/argocd-operator/pull/1019/files

jgwest (changes requested):

go.mod
Mangaal: Hi @jgwest , the dependency in go.mod was not affected by the CVE(it was already using golang.org/x/net v0.24.0), some packages in go.sum were referencing an older, vulnerable version. Running go mod tidy alone did not update these references. By adding the replace directive for golang.org/x/net, ensure all dependencies use the  v0.24.0 version, and go mod tidy then correctly updates go.sum.

jgwest: According to this https://bugzilla.redhat.com/show_bug.cgi?id=2161274, CVE-2022-41717 is fixed by using a newer version of go, rather than an upgrade of the net package

Mangaal:
Closing this PR, Since the packages in the go.mod are not vulnerable anymore.

PR #758, opened 2024-08-07
feat: add support for sidecar container in controller and server

iam-veeramalla:
**What type of PR is this?**
> /kind enhancement
> /kind documentation

Adds support for Sidecar container in Controller and Server

Adds support for Sidecar container in Controller and Server

- Added the E2E tests to validate the feature.
`1-106_validate_appcontroller_initcontainers`
`1-107_validate_server_initcontainers`
`1-108_validate_appcontroller_sidecar`
`1-109_validate_server_sidecar`


Makefile
```diff
@@ -123,6 +123,18 @@ test-e2e: manifests generate fmt vet ## Run e2e tests.
...
+operator-sdk: ## Download operator-sdk locally if necessary.
+ifeq (,$(wildcard $(OPERATOR_SDK)))
+	@{ \
+	set -e ;\
+	mkdir -p $(dir $(OPERATOR_SDK)) ;\
+	OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
+	curl -sSLo $(OPERATOR_SDK) https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk_$${OS}_$${ARCH} ;\
+	chmod +x $(OPERATOR_SDK) ;\
+	}
+endif
```
svghadi, line 137: Is this intentional?

iam-veeramalla: Yes @svghadi

test/openshift/e2e/parallel/1-019_validate_volume_mounts/01-assert.yaml
```diff
@@ -1,3 +1,7 @@
+apiVersion: kuttl.dev/v1beta1
```
varshab1210, line 1: @iam-veeramalla usual timeout for parallel tests is 1200s. By adding a timeout in this file, it'll be overridden. Is this intentional?

anandf: I think, this test is failing due to mismatch in the order of the volume mounts.

anandf:
if the side car containers are dynamically injected by other operators (say istio), gitops-operator should ignore such updates right ? Are we handing that use case ?

svghadi:
> if the side car containers are dynamically injected by other operators (say istio), gitops-operator should ignore such updates right ? Are we handing that use case ?

I don't think we have considered such use-cases where external apps will inject sidecar into operator.

varshab1210:
` 1-031_validate_toolchain/1-check | ArgoCD version mismatch. Should be v2.11.2+25f7504, is v2.11.6+089247d`

@iam-veeramalla could you please update the toolchain test

test/openshift/e2e/parallel/1-019_validate_volume_mounts/01-assert.yaml
varshab1210: Yeah, the test is passing now. We can probably get rid of the timeout

iam-veeramalla:
> > if the side car containers are dynamically injected by other operators (say istio), gitops-operator should ignore such updates right ? Are we handing that use case ?
> 
> I don't think we have considered such use-cases where external apps will inject sidecar into operator.

That's a very good point @anandf , Thank you. However, that change is required in Argo CD Operator. This repository should not be impacted by the changes you suggested.

The API changes, Tests and Manifests changes can be merged.

Once I get the suggested change in to the Argo CD Operator, we can just update the go.mod in this repository to pick the latest Argo CD Operator changes.

PR #771, opened 2024-08-26
Upgraded ArgoCD operator, ArgoCD to pick up the fix error when route api is not available (cherry-pick)

anandf:
**What type of PR is this?**
 /kind bug

Upgraded ArgoCD Operator and ArgoCD to pick up the fix related to ArgoCD UI broken when the route API becomes unavailable momentarily.

When there is an issue with the API server pods, or the API server pods are just coming up after a restart, the current check for API availability done with a discovery client will fail. This makes the operator code to assume that the server does not support the Route API. To fix this, whenever the API availability check fails, the operator checks if the API is registered. If registered, then the operator assumes that the Route API is supported in the cluster and would eventually become available.

Fixes https://issues.redhat.com/browse/GITOPS-4358

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test


svghadi (approved):
/lgtm
/approve

PR #781, opened 2024-10-01
chore: adding new environment variable for rollouts

jparsai:
**What type of PR is this?**

/kind enhancement

This PR is to set value for environment variable CLUSTER_SCOPED_ARGO_ROLLOUTS_NAMESPACES for rollouts controller while installing operator. This value will be used by rollouts manager repo. Here is the PR that will use values provided https://github.com/argoproj-labs/argo-rollouts-manager/pull/95


jgwest (approved):
LGTM, thanks @jparsai!

PR #783, opened 2024-10-04
fix: Remove resource limits for manager container

svghadi:
**What type of PR is this?**
/kind bug

During the operator-sdk upgrade, we set resource limits for the manager container. However, these limits are causing issues for users, particularly on clusters with a large number of secrets and configmaps, as the operator requires significantly more memory to function, leading to OOM kills. We are currently exploring ways to reduce memory consumption, however, this involves significant effort. Therefore, this PR removes the limits to minimize the impact.

Fixes [GITOPS-5665](https://issues.redhat.com/browse/GITOPS-5665)

I have tested this change using the reproducer from JIRA. The operator was functioning properly, consuming over 2GB of memory without any OOM kills.


ishitasequeira (approved):
Lgtm!!

anandf (approved):
LGTM

iam-veeramalla:
LGTM, Thanks @svghadi

PR #802, opened 2024-11-01
Fix race condition in e2e test 1-052_validate_rolebinding_number

jopit:
**What type of PR is this?**

> /kind failing-test

Fixes race condition in e2e test 1-052_validate_rolebinding_number. As per Jonathan's request:

- '02-check_rolebindings.yaml' should loop wait, rather than only running once then failing:
  - first it should loop, waiting for the expected number of rolebindings exist (e.g. similar to an 'Eventually' in ginkgo)
  - then it should loop, waiting that the expected number of rolebindings continue to exists for 20 seconds (similar to 'Consistently' in ginkgo)


jgwest:
Cluster init failure

/retest

jgwest (approved):
LGTM, thanks @jopit!

PR #803, opened 2024-11-06
chore: Upgrade argocd-operator versions

ishitasequeira:
**What type of PR is this?**
/kind chore

Upgrade argocd-operator versions

**Test acceptance criteria**:

* [x] Unit Test
* [x] E2E Test

### DIFF
--- a/test/openshift/e2e/parallel/1-031_validate_toolchain/01-check.yaml
+++ b/test/openshift/e2e/parallel/1-031_validate_toolchain/01-check.yaml
@@ -8,7 +8,7 @@ commands:
       # These variables need to be maintained according to the component matrix: https://spaces.redhat.com/display/GITOPS/GitOps+Component+Matrix
       expected_kustomizeVersion='v5.4.2'
       expected_helmVersion='v3.15.2'
-      expected_argocdVersion='v2.12.6+4dab5bd'
+      expected_argocdVersion='v2.12.7'
       
       if CI="prow"; then
       # when running against openshift-ci
--- a/bundle/manifests/argoproj.io_applications.yaml
+++ b/bundle/manifests/argoproj.io_applications.yaml
@@ -115,6 +115,11 @@ spec:
               sync:
                 description: Sync contains parameters for the operation
                 properties:
+                  autoHealAttemptsCount:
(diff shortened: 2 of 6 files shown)


svghadi:
Can you also update the CRDs here?
https://github.com/redhat-developer/gitops-operator/tree/master/config/crd/bases

test/openshift/e2e/parallel/1-031_validate_toolchain/01-check.yaml
```diff
@@ -8,7 +8,7 @@ commands:
       # These variables need to be maintained according to the component matrix: https://spaces.redhat.com/display/GITOPS/GitOps+Component+Matrix
       expected_kustomizeVersion='v5.4.2'
       expected_helmVersion='v3.15.2'
-      expected_argocdVersion='v2.12.6+4dab5bd'
+      expected_argocdVersion='v2.12.7'
```
anandf, line 11: There should be a short commit id too added isn't it ?

ishitasequeira: This needs to be matched to the version in `go.mod` which is automatically updated when running `go mod download && go mod tidy && go mod vendor`

svghadi:
> @svghadi , the CRDs seem to be already updated. That is why no change came in the CRDs.

@ishitasequeira - Can you confirm this again? I believe there is a CRD change. In your https://github.com/argoproj-labs/argocd-operator/pull/1591 PR, I see a new `autoHealAttemptsCount` field in Application CRD which is not present in `v1.14` branch in this repo.

test/openshift/e2e/parallel/1-031_validate_toolchain/01-check.yaml
anandf: I think the test will fail as it will try to match `v2.12.7`

ishitasequeira: The test seems to be succeeding. The version in `go.mod` is also reflecting `v2.12.7` matching the value in test.

anandf (approved):
LGTM

anandf (approved):
LGTM

PR #808, opened 2024-11-20
Create downstream container image for konflux

anandrkskd:
**What type of PR is this?**

/kind enhancement

This PR adds separate container image for downstream specific builds. 

Fixes [GITOPS-5882](https://issues.redhat.com/browse/GITOPS-5882)

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test

### DIFF
--- a/.tekton/gitops-operator-pull-request.yaml
+++ b/.tekton/gitops-operator-pull-request.yaml
@@ -18,6 +18,12 @@ metadata:
   namespace: rh-openshift-gitops-tenant
 spec:
   params:
+  - name: build-platforms
+    value:
+      - linux/x86_64
+      - linux/arm64
+      - linux/ppc64le
+      - linux/s390x
   - name: git-url
     value: '{{source_url}}'
   - name: revision
--- a/.tekton/gitops-operator-pull-request.yaml
+++ b/.tekton/gitops-operator-pull-request.yaml
@@ -21,9 +21,6 @@ spec:
   - name: build-platforms
     value:
       - linux/x86_64
-      - linux/arm64
-      - linux/ppc64le
-      - linux/s390x
   - name: git-url
     value: '{{source_url}}'
   - name: revision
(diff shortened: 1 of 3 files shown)


Containerfile.plugin
```diff
@@ -0,0 +1,68 @@
...
+
+LABEL \
+    name="openshift-gitops-1/gitops-rhel8-operator" \
+    version=${CI_CONTAINER_VERSION} \
+    License="Apache 2.0" \
+    com.redhat.component="openshift-gitops-operator-container" \
+    com.redhat.delivery.appregistry="false" \
+    release=${CI_CONTAINER_RELEASE} \
+    upstream-version=${CI_UPSTREAM_VERSION} \
+    upstream-vcs-ref="${CI_GITOPS_OPERATOR_UPSTREAM_COMMIT}" \
```
svghadi, line 59: Same as https://github.com/redhat-developer/gitops-backend/pull/47#discussion_r1850019207
> I think we need to find replacements for `CI_*` env vars. If I am not wrong, these vars are CPAAS specific. We need to figure out konflux equivalent for these or manually define them.

svghadi: We can address this during release pipeline configurations, shouldn't be a blocker for merging this PR.

svghadi:
@anandrkskd - Can you enable multi-platform builds? You need to update `build-platforms` parma in tekton configs. 

```yaml
  - name: build-platforms
    value:
    - linux/x86_64
    - linux/arm64
    - linux/ppc64le
    - linux/s390x
```

.tekton/gitops-operator-pull-request.yaml
```diff
@@ -18,6 +18,12 @@ metadata:
   namespace: rh-openshift-gitops-tenant
 spec:
   params:
+  - name: build-platforms
+    value:
+      - linux/x86_64
+      - linux/arm64
+      - linux/ppc64le
+      - linux/s390x
```
svghadi, line 26:
```suggestion
```
Let's run PR builds only on amd64. I noticed long delays in setting up build env for non amd archs.

anandrkskd: Thanks for bringing this up, I was thinking of the same thing, It would be better we use amd64 for PRs.

PR #816, opened 2024-12-02
Fix cluster-scoped runs within gitops-operator Rollouts E2E test script

jgwest:
**What type of PR is this?**

/kind failing-test

As of latest release of Argo Rollouts, we now require CLUSTER_SCOPED_ARGO_ROLLOUTS_NAMESPACES to be set if user is installing outside the openshift-gitops Namespace. 

This requirement broke the Rollouts E2E tests when running via the `run-rollouts-e2e-tests.sh` script in gitops-operator.

The fix is to add/remove the appropriate value from the Subscription (or Deployment).

**Code reviewer guidelines**:

If you install OpenShift GitOps via subscription on clusterbot-style cluster, then run the script, it should pass. 

The other mode (run via CI) isn't supported at the moment, as the Rollouts tests are not yet running as part of OpenShift CI (but hopefully soon!)

### DIFF
--- a/scripts/run-rollouts-e2e-tests.sh
+++ b/scripts/run-rollouts-e2e-tests.sh
@@ -1,10 +1,155 @@
 #!/bin/bash
 
 # The goal of this script is to run the Argo Rollouts operator tests from the argo-rollouts-manager repo against gitops-operator:
-# - Runs the (cluster-scoped) E2E tests of the Argo Rollouts operator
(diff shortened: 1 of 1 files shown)


jgwest:
One cluster init error, and one test flake in '1-069_validate_redis_secure_comm_autotls_ha' which I will submit a fix for.

/retest

jgwest:
A flake in `1-045_validate_repo_exec_timeout` which I will have a separate PR for

PR #824, opened 2025-01-16
Fix intermittent failures of E2E tests 1-045 and 1-069

jgwest:
**What type of PR is this?**
/kind failing-test

- Fixes intermittent failures of E2E tests 1-045 and 1-069
- These tests should not just assume that the value will be updated instantly; they should instead wait for the value to be updated, and fail if it hasn't after some period of time.


jgwest:
/retest
Known intermittent failure and CI failure

PR #862, opened 2025-03-11
Downgrade v1.16 to Argo Rollouts Manager v0.0.5

jgwest:
**What type of PR is this?**
/kind cleanup

This PR re-targets GitOps Operator v1.16 to use Argo Rollouts Manager v0.0.5 (the most recent stable release of Argo Rollouts Manager). The corresponding Argo Rollouts version is v1.7.2.


anandf (approved):
LGTM

PR #876, opened 2025-04-08
Switch to quay.io repo for nginx-unprivileged image to avoid docker hub rate limiting

jgwest:
**What type of PR is this?**
/kind bug

- At present our E2E tests are using `nginxinc/nginx-unprivileged` from Docker Hub. Unfortunately, Docker Hub has strict rate-limiting for unauthenticated users.
- Fortunately, the nginx open source project maintains `nginx-unprivileged` image on quay.io.
- I've verified that the E2E tests still pass with new image location.

### DIFF
--- a/test/examples/nginx/nginx.yaml
+++ b/test/examples/nginx/nginx.yaml
@@ -14,6 +14,6 @@ spec:
     spec:
       containers:
       - name: nginx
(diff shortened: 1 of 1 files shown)


jgwest:
/retest 
Cluster failure

PR #877, opened 2025-04-09
Port Kuttl E2E tests to Ginkgo

jgwest:
I have ported the gitops-operator kuttl sequential/parallel E2E tests to Ginkgo.

It was a very careful translation, maintaining all existing behaviour and test names. This should make it easy to verify that the translation was completed successfully.

I have also created test fixtures (utility functions) which makes it easy to write flexible, maintainable, reliable tests.

This work includes:
- I've added detailed descriptions of what the test is testing, plus individual test step comments in the tests themselves
- Fixed intermittent race conditions when I found them, including removing sleep statements from tests (and replacing them with Eventually/Consistently)
- Audited what each tests does to make sure it makes sense
- Tests are still fast:
	- Sequential: (JGW: need final measurement)
	- Parallel: (JGW: need final measurement)
- Environment variables can be used to customize which tests run:
	- `NON_OLM`: will skip tests that require OLM (this is similar to the non-olm script file, but is built into the test itself)
	- `SKIP_HA_TESTS`: will skip tests that require a cluster with >= 3 nodes
	- `E2E_DEBUG_SKIP_CLEANUP`: will prevent test cleanup at the end of a test, making it easier to debug failures
	- `LOCAL_RUN`: run the tests against gitops-operator running via 'make run'
	- You can specify multiple variables at the same time.
- I've fixed the problem we were seeing in E2E tests: sometimes when certain test fails on GitHub, but there is no log output because they have been flagged as containing sensitive data (and thus removed).
- Kuttl tests were still using v1alpha1, I've switched to use v1beta1 API
- Switched to using quay.io instead of docker hub for some test images, because docker hub causes intermittent failures due to rate limiting.

### A) Run tests against OpenShift GitOps installed via OLM
```		
# 1) Get a cluster from cluster bot, for example, `rosa create 4.17` (for ROSA hypershift cluster with 2 nodes)
# 2) Login and install OpenShift GitOps v1.16 on it via OLM

cd /tmp
git clone git@github.com:jgwest/gitops-operator
cd gitops-operator
git checkout port-e2e-to-ginkgo-feb-2025

# 3a) Run Sequential tests
SKIP_HA_TESTS=true  make e2e-tests-sequential-ginkgo
# If you are using a non-hypershift cluster you can remove SKIP_HA_TESTS var (some of our tests enable redis HA, which require >=3 worker nodes)

# 3b) Run Parallel tests
SKIP_HA_TESTS=true  make e2e-tests-parallel-ginkgo
# As above, can remove SKIP_HA_TESTS if you are on non-hypershift
```

### B) Run E2E tests against local operator (operator running via `make run`)
```
# 1) Start operator locally
CLUSTER_SCOPED_ARGO_ROLLOUTS_NAMESPACES=argo-rollouts,test-rom-ns-1,rom-ns-1,openshift-gitops  make run 
# Add 'ARGOCD_REDIS_IMAGE=quay.io/fedora/redis-7@sha256:0217e8a80a03c3e43c716a1c137e5e615c6bc583605a754000f6efa382349a5b' if you hit docker hub rate-limiting

# 2) Start tests in LOCAL_RUN mode (this skips tests that require Subscription or CSVs)
LOCAL_RUN=true  make e2e-tests-sequential-ginkgo
LOCAL_RUN=true  make e2e-tests-parallel-ginkgo
# Not all tests are supported when running locally. See 'Skip' messages for details.
```

### C) Run a specific test:
```
# 'make ginkgo' to download ginkgo, if needed
# Examples:
./bin/ginkgo -v -focus "1-106_validate_argocd_metrics_controller"  -r ./test/openshift/e2e/ginkgo/sequential
./bin/ginkgo -v -focus "1-099_validate_server_autoscale"  -r ./test/openshift/e2e/ginkgo/parallel
```

## Notes

Additional Notes:
- In parallel, there are 2 '1-063' tests 
- I did not port 'sequential 1-077_validate_workload_status_monitoring_alert' because it is commented out, and has been for the last 2 years
- I did not port 'sequential 1-084_validate_prune_templates' because it is disabled, and the fxnality is not used anymore?
- I did not port 'parallel 1-098_validate_dex_clientsecret' because it has the exact same content as 'parallel 1-095_validate_dex_clientsecret'
- 'parallel 1-058_validate_prometheus_rule' has the name of 'validate prometheus rule', but it doesn't actually do anything with prometheus?
- These tests are up to date with `master` branch kuttl tests as of April 4th, 2025.

Blocked:
- I could not port 'sequential 1-041_validate_argocd_sync_alert' because we're using an extremely old version of prometheus operator PR. We need to fix this first.

I've not moved any tests to/from parallel/sequential, but...

## Recommendations:

I recommend moving these tests from parallel to sequential:
- 'parallel 1-052_validate_rolebinding_number': Since this test indirectly touches openshift-gitops, arguably it should not be parallel. It might break other parallel tests that run at the same time.
- 'parallel 1-074_validate_terminating_namespace_block': Since this test is specifically designed to see if the entire operator can be blocked, arguably it should not be in parallel.
- 'parallel 1-071_validate_SCC_HA': This creates a cluster-scoped resource, and thus should probably be sequential
- 'parallel 1-102_validate_handle_terminating_namespaces':  Since this test is specifically designed to see if the entire operator can be blocked, arguably it should not be in parallel.
- 'parallel 1-105_validate_default_argocd_route': This references and modifies 'openshift-gitops' namespace resources, and thus should not be in parallel.

I recommend moving these tests from sequential to parallel:
- 'sequential 1-050_validate_sso': AFAICT this test doesn't touch any cluster-scoped resources, and thus is a good candidate to run in parallel.
- 'sequential 1-107_validate_redis_ssc': AFAICT this test doesn't touch any cluster-scoped resources, and thus is a good candidate to run in parallel.

(I've not acted on these recommendations as part of this PR, in order to reduce friction for merging the PR)


test/openshift/e2e/ginkgo/parallel/1-007_validate_namespace_scoped_install_test.go
```diff
@@ -0,0 +1,104 @@
+/*
+Copyright 2021.
```
anandf, line 2: year must be`2025`

AdamSaleh:
@jgwest - I am looking at some of the failing tests and it seems we are hitting some limits when downloading things from github when buildint the docker image:

```
STEP 8/11: RUN wget https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh &&     bash install_kustomize.sh /usr/local/bin && rm install_kustomize.sh
--2025-05-19 05:11:02--  https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 185.199.109.133, 185.199.110.133, 185.199.111.133, ...
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|185.199.109.133|:443... connected.
HTTP request sent, awaiting response... 429 Too Many Requests
2025-05-19 05:11:02 ERROR 429: Too Many Requests.
```

Not sure if it is worth it to work around that issue in this PR, and I am not sure if this is generic problem that we are not seeing elsewhere mostly because we don't run e2e tests on i.e. dependabot PRs, but if this will be preventing us from running the e2e tests in ci, we at least need a workaround. Maybe vendoring in that shell-script?

AdamSaleh:
I am currently checking the test-runs locally, mostly to check if there is some if I can see something interesting with regards to code-coverage (I am building the binary with -cover and cunning it with GOCOVERDIR env-var set)

I am running against standard 3+3 node cluster in openstack.

Several of the tests exhibited behavior where they fail in the beforeEach step on ensuring default argocd is runnig - but will work after rerun or after cleaning the cluster and re-running the test. Not sure yet if this is problem with cleanup, or something else, or if this is due to the local environment. This also caused tests running with single test run that would have otherwise been skipped o fail.

In the future I might try running the locally running operator through something like https://telepresence.io/ which could get it really close the regular run, but that is for future experiments.

Only parallel test that didn't work for me was 1-055_drop_resource_customizations and as it tries to call a service api endpoint directly this might just need to be skipped on local runs. 

For sequential tests:

For 1-064_validate_tcp_reset_error_test I got `namespaces "test-1-27-custom" not found` but I don't see it actually being created anywhere, do we expec that sequential test 1-27-_validate_operand_from_git was ran before? 

For 1-083_validate_apps_in_any_namespace I got `HaveReplicas: expected:  1 actual:  2` - as far as I checked, this seems to be valid for multinode cluster. But I will doublechek if there isn't a problem with replicas not spinnning down to correct number.

jgwest:
Hey @AdamSaleh, 

> Several of the tests exhibited behavior where they fail in the beforeEach step on ensuring default argocd is runnig - but will work after rerun or after cleaning the cluster and re-running the test. Not sure yet if this is problem with cleanup, or something else, or if this is due to the local environment. This also caused tests running with single test run that would have otherwise been skipped o fail.

This shouldn't happen: `EnsureCleanSlate` should always ensure that a test is running on as 'clean' or 'stock' and environment as possible (for sequential tests after every test, for parallel, at the beginning of the suite).

If you find cases where this happens, feel free to share the logs: it is probably the case that a previous test failed in an unexpected way, and we need to clean that up by adding additional logic to EnsureCleanSlate. 

(But this is one of the good things about using ginkgo is we CAN use custom logic to find and clean these up between runs.)

> Only parallel test that didn't work for me was 1-055_drop_resource_customizations and as it tries to call a service api endpoint directly this might just need to be skipped on local runs.

This one worked for me on local run (and when run from PR). I looked at what the test is doing and nothing jumped out at me as requiring a non-local env. If you see it again, and have logs, let me know!


> For 1-064_validate_tcp_reset_error_test I got namespaces "test-1-27-custom" not found but I don't see it actually being created anywhere, do we expec that sequential test 1-27-_validate_operand_from_git was ran before?

The 'test-1-27-custom' namespace is deployed by Argo CD via the 'creating Argo CD Application in openshift-gitops namespace' step.

When this test was written, they presumably just copied what 1-027 was doing. Basically only the last step is different.  I  just ported the test as is :grin:.


> For 1-083_validate_apps_in_any_namespace I got HaveReplicas: expected:  1 actual:  2 - as far as I checked, this seems to be valid for multinode cluster. But I will doublechek if there isn't a problem with replicas not spinnning down to correct number.

If you see this again, feel free to share logs, and we can decide if this is a test issue (need to wait longer for replicas to spin down) or operator issue (replicas aren't set correctly).

AdamSaleh:
I reviewed it again and ran it against a proper 3x3 cluster. Looks good. I would say, let's merge it.

jgwest:
Thanks for reviewing @AdamSaleh and @anandf!

PR #927, opened 2025-07-23
Enable gosec/golangci-lint, then fix reported errors

jgwest:
**What type of PR is this?**

/kind cleanup

- Add `gosec` and `golangci-lint` targets to `Makefile`
- Add call to `gosec` and `golangci-lint` (with default settings) to GitHub workflow
- Adds ginkgo-linter, which ensures that Ginkgo code is properly formatted (we used this on managed-gitops project, and it was created by a Red Hatter who reached out to let us know about it, previously)
- The rest of the PR is cleaning up issues identified by `gosec` and `golangci-lint`

- You'll occasionally see cases where I've had to 'smother' a line with lint ignores, like so:
```go
	//lint:ignore SA1019 known to be deprecated
	assert.DeepEqual(t, testArgoCD.Spec.Grafana.Resources, testGrafanaResources) //nolint:staticcheck // SA1019: We must test deprecated fields.
```
- Unfortunately this is necessary to add ignores for BOTH VS Code warnings AND golangci-lint errors


jgwest:
Thanks @jannfis!

Merging, as the  Konflux workflow error appears unrelated to this PR:
```
✕ [Violation] slsa_source_correlated.expected_source_code_reference
  ImageRef: quay.io/redhat-user-workloads/rh-openshift-gitops-tenant/dex@sha256:11a13a97b4450b01934a7632692511d1fef1357af9cdbdfdf29f19854203ac6b
  Reason: The expected source code reference
  "git+https://github.com/rh-gitops-midstream/downstream-dex@149705e191549b57c62b0567a87f99406e96e6ae" is not attested
```

PR #941, opened 2025-08-07
Update to latest commit of argo-rollouts-manager 'a100fc1d88b8e3eee5d0d31a8be2956f96005d77' and argocd-operator '5135028d69782184553b57088981c72125aa77d7'

jgwest:
Update to most recent 'argo-rollouts-manager' commit: https://github.com/argoproj-labs/argo-rollouts-manager/commit/a100fc1d88b8e3eee5d0d31a8be2956f96005d77

Update to the most recent ''argocd-operator' commit: https://github.com/argoproj-labs/argocd-operator/commit/5135028d69782184553b57088981c72125aa77d7

Some additional gitops-operator changes were required due to Namespace Management feature added to argocd-operator. I made them as part of this PR.


cmd/main.go
```diff
@@ -300,3 +308,19 @@ func registerComponentOrExit(mgr manager.Manager, f func(*k8sruntime.Scheme) err
 	}
 	setupLog.Info(fmt.Sprintf("Component registered: %v", reflect.ValueOf(f)))
 }
+
+func initK8sClient() (*kubernetes.Clientset, error) {
```
chetan-rns, line 312: Nit: We could move this function to a different package so that it can be reused everywhere. There's already a similar function [here](https://github.com/redhat-developer/gitops-operator/blob/2f4984c379223e26fb666d5a25517d0e92bf560e/controllers/argocd/openshift/openshift.go#L377).

chetan-rns (approved):
Thanks!

/lgtm

cmd/main.go
jgwest: Agreed, thanks @chetan-rns! I will follow up with this on a separate PR.

jgwest:
Thanks for reviewing @chetan-rns!

PR #948, opened 2025-08-18
Add e2e tests for namespaceManagement

Rizwana777:
JIRA - https://issues.redhat.com/browse/GITOPS-7366

### DIFF
--- a/test/openshift/e2e/ginkgo/sequential/1-113_validate_namespacemanagement_test.go
+++ b/test/openshift/e2e/ginkgo/sequential/1-113_validate_namespacemanagement_test.go
@@ -208,7 +208,7 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 			verifyNamespaceManagementSecretAndApplication(argoCD, nsCustom, true, []string{argoCD.Namespace, nsCustom.Name})
 		})
 
-		It("should not create roles when namespaceManagement env var is not set", func() {
+		It("should not create Roles/RoleBindings when namespaceManagement env var is not set", func() {
 			if fixture.EnvLocalRun() {
 				Skip("This test modifies the Subscription/operator deployment env vars, which requires the operator be running on the cluster.")
 				return
@@ -218,9 +218,6 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 			argoCD := deployArgoCD(randomNS.Name, []argov1beta1api.ManagedNamespaces{
 				{Name: nsCustom.Name, AllowManagedBy: true},
 			})
-
-			By("Enabling namespaceManagement via env var")
-			fixture.SetEnvInOperatorSubscriptionOrDeployment("ALLOW_NAMESPACE_MANAGEMENT_IN_NAMESPACE_SCOPED_INSTANCES", "false")
 
 			By("Create namespaceManagement CR with the namespace which needs to be managed")
 			nm := argov1beta1api.NamespaceManagement{
@@ -236,7 +233,7 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 			verifyNamespaceManagementSecretAndApplication(argoCD, nsCustom, false, []string{})
 		})
 
-		It("should delete Roles/RoleBindings when namespace management is disabled from ArgoCD NamespaceManagement field", func() {
+		It("Should not create Roles/RoleBindings when namespaceManagement is disabled via the ArgoCD NamespaceManagement.AllowManagedBy field.", func() {
 			if fixture.EnvLocalRun() {
 				Skip("This test modifies the Subscription/operator deployment env vars, which requires the operator be running on the cluster.")
 				return
@@ -257,7 +254,7 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 			}
 			Expect(k8sClient.Create(ctx, &nsm)).To(Succeed())
 
-			By("Verify Roles/RoleBindings are deleted from managed namespace")
+			By("Verify Roles/RoleBindings are not created for managed namespace")
 			checkRolesBindings(nsCustom.Name, false)
 
 			By("Verify Application and Secret of managed namespace")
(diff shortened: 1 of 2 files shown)


test/openshift/e2e/ginkgo/sequential/1-113_validate_namespacemanagement_test.go
```diff
@@ -0,0 +1,401 @@
...
+			}
+
+			By("Create ArgoCD with namespaceManagement field set to true")
+			argoCD := deployArgoCD(randomNS.Name, []argov1beta1api.ManagedNamespaces{
+				{Name: nsCustom.Name, AllowManagedBy: true},
+			})
+
+			By("Enabling namespaceManagement via env var")
+			fixture.SetEnvInOperatorSubscriptionOrDeployment("ALLOW_NAMESPACE_MANAGEMENT_IN_NAMESPACE_SCOPED_INSTANCES", "false")
+
```
jgwest: Rather than setting to false, let's just remove this line entirely from this block: we want to verify that if ALLOW_NAMESPACE_MANAGEMENT_IN_NAMESPACE_SCOPED_INSTANCES is not set, that roles/rolebindings are not created.

test/openshift/e2e/ginkgo/sequential/1-113_validate_namespacemanagement_test.go
```diff
@@ -0,0 +1,401 @@
...
+			fixture.SetEnvInOperatorSubscriptionOrDeployment("ALLOW_NAMESPACE_MANAGEMENT_IN_NAMESPACE_SCOPED_INSTANCES", "true")
+
+			By("Create namespaceManagement CR")
+			nsm := argov1beta1api.NamespaceManagement{
+				ObjectMeta: metav1.ObjectMeta{Name: nmName, Namespace: nsCustom.Name},
+				Spec:       argov1beta1api.NamespaceManagementSpec{ManagedBy: randomNS.Name},
+			}
+			Expect(k8sClient.Create(ctx, &nsm)).To(Succeed())
+
+			By("Verify Roles/RoleBindings are deleted from managed namespace")
```
jgwest: Verify Roles/RoleBindings are deleted from managed namespace ->  Verify Roles/RoleBindings are ~deleted~ not created from managed namespace

test/openshift/e2e/ginkgo/sequential/1-113_validate_namespacemanagement_test.go
```diff
@@ -0,0 +1,401 @@
...
+			Expect(k8sClient.Create(ctx, &nm)).To(Succeed())
+
+			By("Verify Roles/RoleBindings are created for managed namespace")
+			checkRolesBindings(nsCustom.Name, false)
+
+			By("Verify Application and Secret of managed namespace")
+			verifyNamespaceManagementSecretAndApplication(argoCD, nsCustom, false, []string{})
+		})
+
+		It("should delete Roles/RoleBindings when namespace management is disabled from ArgoCD NamespaceManagement field", func() {
```
jgwest: Test name is inaccurate: the test as written checks that the Role/RoleBindings are not created if allowManagedBy is false.

It doesn't check if the Role/RoleBindings were deleted.

test/openshift/e2e/ginkgo/sequential/1-113_validate_namespacemanagement_test.go
```diff
@@ -0,0 +1,401 @@
...
+			By("Enabling namespace management via env var")
+			fixture.SetEnvInOperatorSubscriptionOrDeployment("ALLOW_NAMESPACE_MANAGEMENT_IN_NAMESPACE_SCOPED_INSTANCES", "true")
+
+			By("Create namespaceManagement CR")
+			nsm := argov1beta1api.NamespaceManagement{
+				ObjectMeta: metav1.ObjectMeta{Name: nmName, Namespace: nsCustom.Name},
+				Spec:       argov1beta1api.NamespaceManagementSpec{ManagedBy: randomNS.Name},
+			}
+			Expect(k8sClient.Create(ctx, &nsm)).To(Succeed())
+
```
jgwest, line 284: As part of this test, can you also create a `NamespaceManagement` in another namespace, where that namespace DOESN'T match the glob? That way we can verify that glob is matching some namespaces, and not matching other namespaces.

test/openshift/e2e/ginkgo/sequential/1-113_validate_namespacemanagement_test.go
```diff
@@ -0,0 +1,401 @@
...
+			By("Role/Rolebindings should be created for managed namespace if namespaceManagement CR is present")
+			checkRolesBindings(nsCustom.Name, true)
+
+			By("Remove NamespaceManagement to check cleanup of Roles/Rolebindings")
+			Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(argoCD), argoCD)).To(Succeed())
+			argoCD.Spec.NamespaceManagement = nil
+			Expect(k8sClient.Update(ctx, argoCD)).To(Succeed())
```
jgwest: We shouldn't call `k8sClient.Update` directly in Ginkgo tests: this will lead to race conditions because an `Update` calls will fail if the `.status` field of the resource was updated after `Get` was called.

Instead, you can call `k8sFixture.Update`. You will need to cast the obj. Here is an example from another part of the code

```go
			k8sFixture.Update(nsTest_1_12_custom, func(obj client.Object) {
				nsObj, ok := obj.(*corev1.Namespace)
				Expect(ok).To(BeTrue())
				// (update your object as needed)
			})
```

test/openshift/e2e/ginkgo/sequential/1-113_validate_namespacemanagement_test.go
```diff
@@ -0,0 +1,401 @@
...
+
+			By("Verify Roles/RoleBindings are deleted from managed namespace")
+			Eventually(func() bool {
+				checkRolesBindings(nsCustom.Name, false)
+				return true
+			}, "2m", "5s").Should(BeTrue())
```
jgwest: AFAICT this:
```
			Eventually(func() bool {
				checkRolesBindings(nsCustom.Name, false)
				return true
			}, "2m", "5s").Should(BeTrue())
```

can change to:
```
				checkRolesBindings(nsCustom.Name, false)
```

test/openshift/e2e/ginkgo/sequential/1-113_validate_namespacemanagement_test.go
```diff
@@ -0,0 +1,401 @@
...
+
+			By("Verify Application and Secret of managed namespace when nm is disabled")
+			verifyNamespaceManagementSecretAndApplication(argoCD, nsCustom, false, []string{argoCD.Namespace, nsCustom.Name})
+		})
+
+		It("should not create roles when ArgoCD CR has no managedBy entry", func() {
```
jgwest: The test description isn't accurate to what the test is doing AFAICT

test/openshift/e2e/ginkgo/sequential/1-113_validate_namespacemanagement_test.go
```diff
@@ -0,0 +1,401 @@
...
+			}).Should(BeTrue(), "ALLOW_NAMESPACE_MANAGEMENT_IN_NAMESPACE_SCOPED_INSTANCES should be false")
+
+			By("Verify Roles/RoleBindings are deleted for managed namespace when env var is false")
+			checkRolesBindings(nsCustom.Name, false)
+
+			By("Verify Application should not be able to sync to managed namespace when env var is false")
+			verifyNamespaceManagementSecretAndApplication(argoCD, nsCustom, false, []string{argoCD.Namespace, nsCustom.Name})
+		})
+	})
+})
```
jgwest, line 472: Can you also add a test (if not already present):
- Enable NM feature
- Create ArgoCD and NamespaceManagement, with argo cd managing the namespacemanagement
- Verify Role/RoleBindings are created in namespace
- set `AllowManagedBy` to false on ArgoCD CR
- Verify that the roles/rolebindings are removed from namespace

jgwest (changes requested):
I've also pushed a commit to your PR that adds some additional `Consistently` checks alongside the `Eventually` checks, to detect race conditions.

test/openshift/e2e/ginkgo/sequential/1-113_validate_namespacemanagement_test.go
Rizwana777: I have written 2 seperate tests related to this one for AllowmanagedBy: true and another for AllowmanagedBy: false, Now i have added one more test which will toggle between true and false - "should create Roles/RoleBindings when AllowManagedBy is true and remove them when false"

PR #960, opened 2025-09-09
[GITOPS-7475]: Default argocd instance does not get recreated after deletion

akhilnittala:
**What type of PR is this?**

/kind bug

Before these changes, deleting the default ArgoCD instance (openshift-gitops) did not trigger its automatic recreation. However, with these changes, the default instance is now correctly recreated when it is deleted.

No documentation required.

https://issues.redhat.com/browse/GITOPS-7475

Fixes #?

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test

Deploy the GitOps Operator, then delete the default openshift-gitops instance from the openshift-gitops namespace. Verify if the instance is automatically recreated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,14 @@
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= ""
 
+
+# Try to detect Docker or Podman
+CONTAINER_RUNTIME := $(shell command -v docker 2> /dev/null || command -v podman 2> /dev/null)
+
+# If neither Docker nor Podman is found, print an error message and exit
+ifeq ($(CONTAINER_RUNTIME),)
+$(warning "No container runtime (Docker or Podman) found in PATH. Please install one of them.")
+endif
 # CHANNELS define the bundle channels used in the bundle.
 # Add a new line here if you would like to change its default config. (E.g CHANNELS = "preview,fast,stable")
 # To re-generate a bundle for other specific channels without changing the standard setup, you can:
@@ -206,12 +214,12 @@ run: manifests generate fmt vet ## Run a controller from your host.
 	CLUSTER_SCOPED_ARGO_ROLLOUTS_NAMESPACES=argo-rollouts,test-rom-ns-1,rom-ns-1,openshift-gitops  ARGOCD_CLUSTER_CONFIG_NAMESPACES=openshift-gitops  REDIS_CONFIG_PATH="build/redis"   go run ./cmd/main.go
 
 .PHONY: docker-build
-docker-build: test ## Build docker image with the manager.
-	docker build -t ${IMG} .
+docker-build: test ## Build $(CONTAINER_RUNTIME) image with the manager.
+	$(CONTAINER_RUNTIME) build -t ${IMG} .
 
 .PHONY: docker-push
-docker-push: ## Push docker image with the manager.
-	docker push ${IMG}
+docker-push: ## Push $(CONTAINER_RUNTIME) image with the manager.
+	$(CONTAINER_RUNTIME) push ${IMG}
 
 ##@ Build Dependencies
 
@@ -305,11 +313,11 @@ bundle: operator-sdk manifests kustomize ## Generate bundle manifests and metada
 
 .PHONY: bundle-build
 bundle-build: ## Build the bundle image.
-	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
+	$(CONTAINER_RUNTIME) build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
 
 .PHONY: bundle-push
 bundle-push: ## Push the bundle image.
-	$(MAKE) docker-push IMG=$(BUNDLE_IMG)
+	$(MAKE) $(CONTAINER_RUNTIME)-push IMG=$(BUNDLE_IMG)
 
 .PHONY: opm
 OPM = ./bin/opm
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -27,7 +35,7 @@ endif
 BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
 # IMAGE defines the base image to be used for operator, bundle and catalog.
-IMAGE ?= quay.io/redhat-developer/gitops-operator
+IMAGE ?= quay.io/nittalaakhil/openshift-gitops-operator
```
anandf: This needs to be reverted. For your testing, you can try setting the variable in environment. Like `IMAGE=quay.io/nittalaakhil/openshift-gitops-operator make build`

controllers/gitopsservice_controller.go
```diff
@@ -112,7 +112,11 @@ func (r *ReconcileGitopsService) SetupWithManager(mgr ctrl.Manager) error {
 			builder.WithPredicates(predicate.NewPredicateFuncs(func(obj client.Object) bool {
 				return obj.GetName() == "openshift-gitops"
 			})),
-		).
+		).Watches(&argoapp.ArgoCD{},
+		&handler.EnqueueRequestForObject{},
+		builder.WithPredicates(predicate.NewPredicateFuncs(func(obj client.Object) bool {
+			return obj.GetName() == "openshift-gitops"
```
anandf: probably we can include namespace too. `obj.GetNamespace() == "openshift-gitops"`

anandf (changes requested):
Thanks @akhilnittala for working on this issue. Posted few comments. Otherwise looks good.

controllers/gitopsservice_controller.go
akhilnittala: sure anand wil add

Makefile
akhilnittala: done, removed

Makefile
```diff
@@ -206,12 +214,12 @@ run: manifests generate fmt vet ## Run a controller from your host.
 	CLUSTER_SCOPED_ARGO_ROLLOUTS_NAMESPACES=argo-rollouts,test-rom-ns-1,rom-ns-1,openshift-gitops  ARGOCD_CLUSTER_CONFIG_NAMESPACES=openshift-gitops  REDIS_CONFIG_PATH="build/redis"   go run ./cmd/main.go
 
 .PHONY: docker-build
-docker-build: test ## Build docker image with the manager.
-	docker build -t ${IMG} .
+docker-build: test ## Build $(CONTAINER_RUNTIME) image with the manager.
```
svghadi:
```suggestion
docker-build: test ## Build container image with the manager.
```

Makefile
```diff
@@ -206,12 +214,12 @@ run: manifests generate fmt vet ## Run a controller from your host.
...
+	$(CONTAINER_RUNTIME) build -t ${IMG} .
 
 .PHONY: docker-push
-docker-push: ## Push docker image with the manager.
-	docker push ${IMG}
+docker-push: ## Push $(CONTAINER_RUNTIME) image with the manager.
```
svghadi:
```suggestion
docker-push: ## Push container image with the manager.
```

Makefile
```diff
@@ -346,12 +354,12 @@ endif
...
 .PHONY: catalog-build
 catalog-build: opm ## Build a catalog image.
-	$(OPM) index add --container-tool docker --mode semver --tag $(CATALOG_IMG) --bundles $(BUNDLE_IMGS) $(FROM_INDEX_OPT)
+	$(OPM) index add --container-tool $(CONTAINER_RUNTIME) --mode semver --tag $(CATALOG_IMG) --bundles $(BUNDLE_IMGS) $(FROM_INDEX_OPT)
 
 # Push the catalog image.
 .PHONY: catalog-push
 catalog-push: ## Push a catalog image.
-	$(MAKE) docker-push IMG=$(CATALOG_IMG)
+	$(MAKE) $(CONTAINER_RUNTIME)-push IMG=$(CATALOG_IMG)
```
svghadi:
```suggestion
	$(MAKE) docker-push IMG=$(CATALOG_IMG)
```
I believe `docker-push` here is referencing to a make target.  If the `$(CONTAINER_RUNTIME)` value is `podman` then the execution will fail with this change as there will be no `podman-push` make target.

akhilnittala:
> Thanks @akhilnittala. Overall looks good. Left some comments.
> 
> Did we run any tests with `DISABLE_DEFAULT_ARGOCD_INSTANCE=true` env? This env var can be used to control creation and deletion of default ArgoCD instance. When set, the default instance should not be created.
> 
> https://docs.redhat.com/en/documentation/red_hat_openshift_gitops/1.13/html/argo_cd_instance/setting-up-argocd-instance

Yes siddhesh, added env variable in manager.yaml in config files as - name: DISABLE_DEFAULT_ARGOCD_INSTANCE ; value: "true" and tested it, it didnt create default argocd instance.

svghadi:
/lgtm

Thanks @akhilnittala . The change looks good. Can we add a sequential e2e test? You can use this as a reference: https://github.com/redhat-developer/gitops-operator/blob/master/test/openshift/e2e/ginkgo/sequential/1-018_validate_disable_default_instance_test.go

test/openshift/e2e/ginkgo/sequential/1-018_validate_disable_default_instance_test.go
```diff
@@ -37,13 +40,71 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 			fixture.EnsureSequentialCleanSlate()
 		})
 
-		It("verifies that DISABLE_DEFAULT_ARGOCD_INSTANCE env var will delete the argo cd instance from openshift-gitops, and that default Argo CD instance will be restored when the env var is removed", func() {
+		It("verifies that ArgoCD instance can be manually deleted and resources are cleaned up", func() {
```
svghadi: The test verifies that if the default ArgoCD instance is deleted manually, the operator recreates it. Can we update the description to reflect that? For example:

```suggestion
		It("verifies that the default ArgoCD instance from openshift-gitops namespace is recreated when deleted manually", func() {
```

PR #962, opened 2025-09-09
added ginkgo test for Banner property of argocd

alkakumari016:
**What type of PR is this?**

> /kind-test

Fixes #?
[GITOPS-7712](https://issues.redhat.com/browse/GITOPS-7712)

### DIFF
--- a/test/e2e/suite_test.go
+++ b/test/e2e/suite_test.go
@@ -171,6 +171,9 @@ var _ = BeforeSuite(func() {
 		Client:    mgr.GetClient(),
 		Scheme:    mgr.GetScheme(),
 		K8sClient: k8sClient,
+		LocalUsers: &argocdprovisioner.LocalUsersInfo{
+			TokenRenewalTimers: map[string]*argocdprovisioner.TokenRenewalTimer{},
+		},
 	}).SetupWithManager(mgr)
 	Expect(err).NotTo(HaveOccurred())
 
--- a/test/nondefaulte2e/suite_test.go
+++ b/test/nondefaulte2e/suite_test.go
@@ -158,6 +158,9 @@ var _ = BeforeSuite(func() {
 		Client:    mgr.GetClient(),
 		Scheme:    mgr.GetScheme(),
 		K8sClient: k8sClient,
+		LocalUsers: &argocdprovisioner.LocalUsersInfo{
+			TokenRenewalTimers: map[string]*argocdprovisioner.TokenRenewalTimer{},
+		},
 	}).SetupWithManager(mgr)
 	Expect(err).NotTo(HaveOccurred())
 


test/openshift/e2e/ginkgo/parallel/1-120_validate_banner_test.go
```diff
@@ -0,0 +1,435 @@
...
+				}, service)
+				if err != nil {
+					GinkgoWriter.Printf("Failed to get service: %v\n", err)
+					return false
+				}
+
+				GinkgoWriter.Printf("Service found: %s\n", service.Name)
+				return true
+			}, "120s", "5s").Should(BeTrue())
+		})
```
svghadi: Why is this required?

test/openshift/e2e/ginkgo/parallel/1-120_validate_banner_test.go
```diff
@@ -0,0 +1,435 @@
...
+
+		AfterEach(func() {
+			fixture.OutputDebugOnFail(ns)
+
+			if cleanupFunc != nil {
+				cleanupFunc()
+			}
+		})
+
+		It("verifies ArgoCD banner configuration is applied and visible in UI", func() {
```
svghadi: I believe the test is performing checks in argocd-cm and not UI. This sentence can be misleading. Can we update it to reflect this?

svghadi (changes requested):
@alkakumari016 - What do you think about combining all the test scenarios in a single IT, instead of creating multiple ITs, which creates multiple ArgoCD instances. 

The test can follow below sequence
1. Create Banner with all support configs in ArgoCD CR
2. Check if it is reflected in argocd-cm
3. Update some banner properties in ArgoCD CR
4. Check if new values are reflected in argocd-cm and old are preserved
5. Remove the banner from ArgoCD CR
6. Check if corresponding configs are removed from argocd-cm

test/openshift/e2e/ginkgo/parallel/1-120_validate_banner_test.go
alkakumari016: Updated

test/openshift/e2e/ginkgo/parallel/1-120_validate_banner_test.go
alkakumari016: It's not required. I have removed it.

alkakumari016:
> @alkakumari016 - What do you think about combining all the test scenarios in a single IT, instead of creating multiple ITs, which creates multiple ArgoCD instances.
> 
> The test can follow below sequence
> 
> 1. Create Banner with all support configs in ArgoCD CR
> 2. Check if it is reflected in argocd-cm
> 3. Update some banner properties in ArgoCD CR
> 4. Check if new values are reflected in argocd-cm and old are preserved
> 5. Remove the banner from ArgoCD CR
> 6. Check if corresponding configs are removed from argocd-cm

@svghadi I have merged the blocks into one It block and removed some unwanted tests. The sequence looks good and makes sense to remove other scenarios since these are covering pretty much everyuthing. I have updated the PR please take a look.

PR #1065, opened 2026-02-04
Port 1-092_validate_workload_status_monitoring_alert-test to Ginkgo

trdoyle81:
**What type of PR is this?**

> /kind enhancement

Port the Kuttl test 1-092_validate_workload_status_monitoring_alert to Ginkgo. The test validates that the operator correctly manages Prometheus monitoring rules and alerts when enabled and verifies that alerts fire when an invalid image configuration causes a workload to degrade.

Moved tthe test to sequential from (kuttl) parallel test suite because it modifies shared cluster-wide resources. Running this test in parallel has caused race conditions and test flakes. 

Fixes #GITOPS-7204

**Test acceptance criteria**:

* [ ] Unit Test
* [x] E2E Test

export GINKGO_TEST_SUITE="sequential"
export GINKGO_FOCUS="1-092_validate_workload_status_monitoring_alert"
./bin/ginkgo -v -focus $GINKGO_FOCUS -r ./test/openshift/e2e/ginkgo/$GINKGO_TEST_SUITE

### DIFF
--- a/test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
+++ b/test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
@@ -1,0 +1,144 @@
+package sequential
+
+import (
+	"context"
+
+	argov1beta1api "github.com/argoproj-labs/argocd-operator/api/v1beta1"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/redhat-developer/gitops-operator/test/openshift/e2e/ginkgo/fixture"
+	argocdFixture "github.com/redhat-developer/gitops-operator/test/openshift/e2e/ginkgo/fixture/argocd"
+	k8sFixture "github.com/redhat-developer/gitops-operator/test/openshift/e2e/ginkgo/fixture/k8s"
+	fixtureUtils "github.com/redhat-developer/gitops-operator/test/openshift/e2e/ginkgo/fixture/utils"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
+
+	Context("1-092_validate_workload_status_monitoring_alert", func() {
+		var (
+			k8sClient    client.Client
+			ctx          context.Context
+			nsCluster    *corev1.Namespace
+			nsNamespaced *corev1.Namespace
+			cleanupFunc  func()
+		)
+
+		BeforeEach(func() {
+			fixture.EnsureSequentialCleanSlate()
+			k8sClient, _ = fixtureUtils.GetE2ETestKubeClient()
+			ctx = context.Background()
+
+			nsCluster = &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "openshift-gitops"}}
+			nsNamespaced, cleanupFunc = fixture.CreateRandomE2ETestNamespaceWithCleanupFunc()
+		})
+
+		AfterEach(func() {
+			defer cleanupFunc()
+			fixture.OutputDebugOnFail(nsNamespaced)
+		})
+
+		It("validates monitoring setup, alert rule creation, and teardown", func() {
+			const (
+				invalidImage        = "test-image"
+				prometheusRuleName  = "gitops-operator-argocd-alerts"
+				clusterInstanceName = "openshift-gitops"
+			)
+
+			ruleCluster := &monitoringv1.PrometheusRule{
+				ObjectMeta: metav1.ObjectMeta{Name: prometheusRuleName, Namespace: nsCluster.Name},
+			}
+			ruleNamespaced := &monitoringv1.PrometheusRule{
+				ObjectMeta: metav1.ObjectMeta{Name: prometheusRuleName, Namespace: nsNamespaced.Name},
+			}
+
+			uwmConfigMap := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{Name: "cluster-monitoring-config", Namespace: "openshift-monitoring"},
+				Data:       map[string]string{"config.yaml": "enableUserWorkload: true"},
+			}
+			cmKey := client.ObjectKey{Name: uwmConfigMap.Name, Namespace: uwmConfigMap.Namespace}
+
+			By("enabling user workload monitoring in the cluster monitoring config map")
+			existingCM := &corev1.ConfigMap{}
+			err := k8sClient.Get(ctx, cmKey, existingCM)
+
+			DeferCleanup(func() {
+				_ = k8sClient.Delete(ctx, uwmConfigMap)
+			})
+
+			if err == nil {
+				existingCM.Data = uwmConfigMap.Data
+				Expect(k8sClient.Update(ctx, existingCM)).To(Succeed(), "Failed to update existing UWM ConfigMap")
+			} else {
+				Expect(k8sClient.Create(ctx, uwmConfigMap)).To(Succeed(), "Failed to create UWM ConfigMap")
+			}
+
+			By("enabling monitoring on the cluster Argo CD instance and setting an invalid image to trigger alerts")
+			argoCDCluster := &argov1beta1api.ArgoCD{
+				ObjectMeta: metav1.ObjectMeta{Name: clusterInstanceName, Namespace: nsCluster.Name},
+			}
+			Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(argoCDCluster), argoCDCluster)).To(Succeed())
+
+			//restore the cluster instance even if the test fails halfway through it
+			DeferCleanup(func() {
+				By("restoring the default image and disabling monitoring on cluster Argo CD instance (Cleanup)")
+				_ = k8sClient.Get(ctx, client.ObjectKeyFromObject(argoCDCluster), argoCDCluster)
+				argocdFixture.Update(argoCDCluster, func(ac *argov1beta1api.ArgoCD) {
+					ac.Spec.ApplicationSet.Image = ""
+					ac.Spec.Monitoring.DisableMetrics = ptr.To(true)
+				})
+			})
+
+			argocdFixture.Update(argoCDCluster, func(ac *argov1beta1api.ArgoCD) {
+				ac.Spec.ApplicationSet = &argov1beta1api.ArgoCDApplicationSet{Image: invalidImage}
+				ac.Spec.Monitoring = argov1beta1api.ArgoCDMonitoringSpec{DisableMetrics: ptr.To(false)}
+			})
+
+			By("creating a namespaced Argo CD instance with monitoring enabled")
+			argoCDNamespaced := &argov1beta1api.ArgoCD{
+				ObjectMeta: metav1.ObjectMeta{Name: "argocd", Namespace: nsNamespaced.Name},
+				Spec: argov1beta1api.ArgoCDSpec{
+					ApplicationSet: &argov1beta1api.ArgoCDApplicationSet{Image: invalidImage},
+					Monitoring:     argov1beta1api.ArgoCDMonitoringSpec{DisableMetrics: ptr.To(false)},
+				},
+			}
+			Expect(k8sClient.Create(ctx, argoCDNamespaced)).To(Succeed())
+
+			//the verification
+			By("waiting for the Argo CD instances to become available")
+			Eventually(argoCDCluster, "5m").Should(argocdFixture.BeAvailable())
+			Eventually(argoCDNamespaced, "5m").Should(argocdFixture.BeAvailable())
+
+			By("verifying the operator created the expected PrometheusRules")
+			Eventually(ruleCluster, "5m").Should(k8sFixture.ExistByName(), "PrometheusRule should be created in cluster namespace")
+			Eventually(ruleNamespaced, "5m").Should(k8sFixture.ExistByName(), "PrometheusRule should be created in test namespace")
+
+			By("verifying the ApplicationSet deployments are present (likely in a crash loop due to the invalid image)")
+			appSetDeplCluster := &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: clusterInstanceName + "-applicationset-controller", Namespace: nsCluster.Name}}
+			Eventually(appSetDeplCluster).Should(k8sFixture.ExistByName())
+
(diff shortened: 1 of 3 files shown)


test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
```diff
@@ -0,0 +1,144 @@
...
+
+		AfterEach(func() {
+			defer cleanupFunc()
+			fixture.OutputDebugOnFail(nsNamespaced)
+		})
+
+		It("validates monitoring setup, alert rule creation, and teardown", func() {
+			const (
+				invalidImage        = "test-image"
+				prometheusRuleName  = "gitops-operator-argocd-alerts"
```
varshab1210: Has the naming convention changed? It used to be `argocd-component-status-alert` according to the kuttl test we have

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
```diff
@@ -0,0 +1,144 @@
...
+			argoCDNamespaced := &argov1beta1api.ArgoCD{
+				ObjectMeta: metav1.ObjectMeta{Name: "argocd", Namespace: nsNamespaced.Name},
+				Spec: argov1beta1api.ArgoCDSpec{
+					ApplicationSet: &argov1beta1api.ArgoCDApplicationSet{Image: invalidImage},
+					Monitoring:     argov1beta1api.ArgoCDMonitoringSpec{DisableMetrics: ptr.To(false)},
+				},
+			}
+			Expect(k8sClient.Create(ctx, argoCDNamespaced)).To(Succeed())
+
+			//the verification
```
varshab1210: Can we add verification step for the alert as well?

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
trdoyle81: Thanks for catching this. I've updated the rule name back to argocd-component-status-alert in the latest commit to make it the same as the original Kuttl test.

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
trdoyle81: Done! I've added an Eventually block that polls the internal Prometheus API to make sure the ApplicationSetControllerNotReady alerts got to the "firing" state for the default and custom instances

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
```diff
@@ -0,0 +1,189 @@
...
+			nsNamespaced.Labels["openshift.io/cluster-monitoring"] = "true"
+			err = k8sClient.Update(ctx, nsNamespaced)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("enabling user workload monitoring in the cluster monitoring config map")
+			existingCM := &corev1.ConfigMap{}
+			err = k8sClient.Get(ctx, client.ObjectKeyFromObject(uwmConfigMap), existingCM)
+
+			DeferCleanup(func() {
+				_ = k8sClient.Delete(ctx, uwmConfigMap)
```
varshab1210: @trdoyle81 uwmConfigMap is something that already exists on the cluster right? Should we just revert the changes that was made instead of deleting it?

Maybe we can move the function below where the creation takes place? That way we can either revert the changes or delete the configmap it was created using the test step

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
```diff
@@ -0,0 +1,189 @@
...
+				argocdFixture.Update(argoCDCluster, func(ac *argov1beta1api.ArgoCD) {
+					ac.Spec.ApplicationSet.Image = ""
+					ac.Spec.Monitoring.DisableMetrics = ptr.To(true)
+					ac.Spec.Monitoring.Enabled = false
+				})
+			})
+
+			argocdFixture.Update(argoCDCluster, func(ac *argov1beta1api.ArgoCD) {
+				ac.Spec.ApplicationSet = &argov1beta1api.ArgoCDApplicationSet{Image: invalidImage}
+				ac.Spec.Monitoring = argov1beta1api.ArgoCDMonitoringSpec{Enabled: true, DisableMetrics: ptr.To(false)}
```
varshab1210: Do we need to set `DisableMetrics`?

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
```diff
@@ -0,0 +1,189 @@
...
+
+			By("disabling monitoring and restoring the default images")
+			Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(argoCDCluster), argoCDCluster)).To(Succeed())
+			argocdFixture.Update(argoCDCluster, func(ac *argov1beta1api.ArgoCD) {
+				ac.Spec.ApplicationSet.Image = ""
+				ac.Spec.Monitoring.Enabled = false
+				ac.Spec.Monitoring.DisableMetrics = ptr.To(true)
+			})
+
+			Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(argoCDNamespaced), argoCDNamespaced)).To(Succeed())
```
varshab1210, line 205: This is not needed IMO as we are going to delete this anyway

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
```diff
@@ -0,0 +1,189 @@
...
+			argocdFixture.Update(argoCDNamespaced, func(ac *argov1beta1api.ArgoCD) {
+				ac.Spec.ApplicationSet.Image = ""
+				ac.Spec.Monitoring.Enabled = false
+				ac.Spec.Monitoring.DisableMetrics = ptr.To(true)
+			})
+
+			By("deleting cluster monitoring config")
```
varshab1210: Same comment as above

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
```diff
@@ -0,0 +1,189 @@
...
+				clusterInstanceName = "openshift-gitops"
+			)
+
+			ruleCluster := &monitoringv1.PrometheusRule{
+				ObjectMeta: metav1.ObjectMeta{Name: prometheusRuleName, Namespace: nsCluster.Name},
+			}
+			ruleNamespaced := &monitoringv1.PrometheusRule{
+				ObjectMeta: metav1.ObjectMeta{Name: prometheusRuleName, Namespace: nsNamespaced.Name},
+			}
+			appSetDeplCluster := &appsv1.Deployment{
```
varshab1210, line 62: This doesn't seem to be required as we are checking the appset directly

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
```diff
@@ -0,0 +1,189 @@
...
+			By("waiting for the Argo CD instances to become available")
+			Eventually(argoCDCluster, "5m").Should(argocdFixture.BeAvailable())
+			Eventually(argoCDNamespaced, "5m").Should(argocdFixture.BeAvailable())
+
+			By("verifying PrometheusRules are created with the correct alerts")
+			Eventually(ruleCluster, "3m", "5s").Should(k8sFixture.ExistByName(), "PrometheusRule should be created in cluster namespace")
+			Eventually(ruleNamespaced, "3m", "5s").Should(k8sFixture.ExistByName(), "PrometheusRule should be created in test namespace")
+
+			By("verifying the ApplicationSet deployments are present")
+			Eventually(appSetDeplCluster).Should(k8sFixture.ExistByName())
```
varshab1210, line 153: We will need to add one for the namespace scoped appset pod as well

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
```diff
@@ -0,0 +1,189 @@
...
+
+			By("verifying the workload degradation alerts are actively firing in Prometheus")
+			Eventually(func() bool {
+				cmd := exec.Command("oc", "exec", "-n", "openshift-monitoring", "prometheus-k8s-0", "-c", "prometheus", "--", "curl", "-s", "http://localhost:9090/api/v1/alerts")
```
varshab1210, line 158: Is there a way to parse is via jq or some other tool as discussed?

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
trdoyle81: Thanks. I've moved the DeferCleanup block below the evaluation so now it checks if the ConfigMap existed beforehand and if it did then it restores it on teardowna and if not gets deleted.

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
trdoyle81: You are right because Enabled = false handles the teardown so this is redundant. I removed DisableMetrics = ptr.To(true) from teardown and cleanup.

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
trdoyle81: Done

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
trdoyle81: Done

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
trdoyle81: I have added appSetDeplNamespaced and added an Eventually for checking that the namespace-scoped deployment exists with the cluster-scoped one.

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
trdoyle81: I kept this and added the namespaced deployment. let me know if you'd prefer I remove both deployment checks.

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
trdoyle81: Done

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
```diff
@@ -98,19 +100,30 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 				Expect(err).NotTo(HaveOccurred(), "Failed to fetch UWM ConfigMap")
 			}
 
+			DeferCleanup(func() {
+				By("restoring or deleting cluster monitoring config")
+				if cmExisted {
+					revertCM := &corev1.ConfigMap{}
+					Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(uwmConfigMap), revertCM)).To(Succeed())
+					revertCM.Data = originalData
```
varshab1210, line 107: This might cause an error since originalData is defined inside the if block

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
```diff
@@ -98,19 +100,30 @@ var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 				Expect(err).NotTo(HaveOccurred(), "Failed to fetch UWM ConfigMap")
 			}
 
+			DeferCleanup(func() {
+				By("restoring or deleting cluster monitoring config")
+				if cmExisted {
+					revertCM := &corev1.ConfigMap{}
+					Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(uwmConfigMap), revertCM)).To(Succeed())
```
varshab1210, line 106: Is it required to fetch the uwmConfigMap if we are going to assign the data from originalData? If its for the metadata field we can initialise the revertCM this way I believe

```
			revertCM := &corev1.ConfigMap{
				ObjectMeta: metav1.ObjectMeta{Name: "cluster-monitoring-config", Namespace: "openshift-monitoring"},
				Data:       map[string]string,
			}
```

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
varshab1210: Can you remove it from other places as well?

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
trdoyle81: I double-checked this and originalData is actually declared outside of the if block (var originalData map[string]string on line 90) so it in scope for the DeferCleanup closure. I'm guessing the compiler would probably complain if this was an issue.

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
trdoyle81: If we use k8sClient.Update() then we'll need to pass it the latest ResourceVersion of the object otherwise it would reject the request with something like a Conflict/Invalid error. Doing a fresh Get() inside the cleanup makes sure we have the current ResourceVersion before we apply the originalData and do the update. If we want to avoid the Get() we'll need to switch to using k8sClient.Patch() but I think Get() and then Update() is probably safer to prevent any conflicts ?

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
trdoyle81: good catch I removed DisableMetrics: ptr.To(false) from the setup phases as well. So now relies on Enabled: true and defaults for the rest.

test/openshift/e2e/ginkgo/sequential/1-092_validate_workload_status_monitoring_alert.go
varshab1210: You are right, I got this confused with the next case

jgwest:
cluster flake
/retest

PR #1092, opened 2026-02-27
Added unit-test codecov upload

AdamSaleh:
**What type of PR is this?**
/kind enhancement

We should be tracking our code coverage, this should enable it in such away that it won't cause test to fail if upload fails.

### DIFF
--- a/scripts/openshiftci-presubmit-unittests.sh
+++ b/scripts/openshiftci-presubmit-unittests.sh
@@ -27,3 +27,16 @@ cd ../..
 
 # Run unit
 make test
+
+# Upload coverage to codecov.io - failures here should not fail the build
+(
+  set +e
+  CODECOV_TOKEN_FILE="/var/run/codecov-token/CODECOV_TOKEN"
+  if [[ ! -f "${CODECOV_TOKEN_FILE}" ]]; then
+    echo "Codecov token not found at ${CODECOV_TOKEN_FILE}, skipping upload"
+    exit 0
+  fi
+  curl -OSs --fail-with-body https://cli.codecov.io/latest/linux/codecov
+  chmod +x codecov
+  CODECOV_TOKEN="$(cat "${CODECOV_TOKEN_FILE}")" ./codecov upload-process --flag unit-tests --file cover.out
+) || echo "Coverage upload to codecov.io failed, continuing"
--- a/scripts/openshiftci-presubmit-unittests.sh
+++ b/scripts/openshiftci-presubmit-unittests.sh
@@ -36,7 +36,22 @@ make test
     echo "Codecov token not found at ${CODECOV_TOKEN_FILE}, skipping upload"
     exit 0
   fi
-  curl -OSs --fail-with-body https://cli.codecov.io/latest/linux/codecov
-  chmod +x codecov
-  CODECOV_TOKEN="$(cat "${CODECOV_TOKEN_FILE}")" ./codecov upload-process --flag unit-tests --file cover.out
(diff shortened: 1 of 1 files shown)


scripts/openshiftci-presubmit-unittests.sh
```diff
@@ -27,3 +27,16 @@ cd ../..
...
+(
+  set +e
+  CODECOV_TOKEN_FILE="/var/run/codecov-token/CODECOV_TOKEN"
+  if [[ ! -f "${CODECOV_TOKEN_FILE}" ]]; then
+    echo "Codecov token not found at ${CODECOV_TOKEN_FILE}, skipping upload"
+    exit 0
+  fi
+  curl -Os https://cli.codecov.io/latest/linux/codecov
+  chmod +x codecov
+  CODECOV_TOKEN="$(cat "${CODECOV_TOKEN_FILE}")" ./codecov upload-process --file cover.out
```
psturc: should we also specify the flag "unit-tests" so we can distinguish between test types in a future once we start uploading e2e test coverage too?

e.g.

```
./codecov upload-process \
  --token "${CODECOV_TOKEN}" \
  --flag unit-tests \
  --file coverage.out
```
see the jira ticket for more details

scripts/openshiftci-presubmit-unittests.sh
```diff
@@ -27,3 +27,16 @@ cd ../..
...
+
+# Upload coverage to codecov.io - failures here should not fail the build
+(
+  set +e
+  CODECOV_TOKEN_FILE="/var/run/codecov-token/CODECOV_TOKEN"
+  if [[ ! -f "${CODECOV_TOKEN_FILE}" ]]; then
+    echo "Codecov token not found at ${CODECOV_TOKEN_FILE}, skipping upload"
+    exit 0
+  fi
+  curl -Os https://cli.codecov.io/latest/linux/codecov
```
olivergondza: nit: prefer options `-Ss --fail-with-body` for easier diagnostics in case it fails in CI.

scripts/openshiftci-presubmit-unittests.sh
AdamSaleh: Hm, fair point. Will add.

ranakan19 (changes requested):
Can we follow the official documentation for binary integrity check as well https://docs.codecov.com/docs/the-codecov-cli#integrity-checking-the-binary? while keeping the version pin intact
from my limited knowledge on the subject, my understanding is sha256sum does file integrity check but does verify the source of the file.

AdamSaleh:
Good idea, ammended

ranakan19 (approved):
Looks good

scripts/openshiftci-presubmit-unittests.sh
varshab1210, line 1: Seems like the token is not mounted, is this expected?

```
+ set +e
+ CODECOV_TOKEN_FILE=/var/run/codecov-token/CODECOV_TOKEN
+ [[ ! -f /var/run/codecov-token/CODECOV_TOKEN ]]
+ echo 'Codecov token not found at /var/run/codecov-token/CODECOV_TOKEN, skipping upload'
Codecov token not found at /var/run/codecov-token/CODECOV_TOKEN, skipping upload
+ exit 0
```

AdamSaleh: It works now! Had to upload it directly through bash, but it works - https://app.codecov.io/github/redhat-developer/gitops-operator/commit/a25055be57a7a7f1b49924435b35c066410b3920

PR #1157, opened 2026-05-25
Remove hard dependency on OCP CRDs  

anandrkskd:
assisted-by: claude-code
**What type of PR is this?**

/kind enhancement

Before this change, Openshift GitOps uses Routes and other Openshift specific tests as hard dependency. This PR removes this hard dependency that willl allow users to run gitops-operator on non OCP clusters. This PR also adds a check to see if promethus (monitoring.coreos.com) is present, if not it skips.

Fixes https://redhat.atlassian.net/browse/GITOPS-9812

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test

- start a any k8s cluster, create argocd namespace 
- run make target
```sh
make install run
cat <<EOF | kubectl apply -f -
apiVersion: argoproj.io/v1beta1
kind: ArgoCD
metadata:
  name: test
  namespace: argocd
spec: {}
EOF
```
- check logs if any openshift api cause failure

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -223,7 +223,7 @@ func main() {
 	registerComponentOrExit(mgr, argov1beta1api.AddToScheme)
 
 	// Setup Scheme for OpenShift Config if available
-	if util.IsOpenShiftCluster() {
+	if util.IsConfigAPIFound() {
 		registerComponentOrExit(mgr, configv1.AddToScheme)
 	}
 
@@ -254,17 +254,13 @@ func main() {
 		}
 	}
 
-	if util.IsOpenShiftCluster() {
-		if err = (&controllers.ReconcileGitopsService{
-			Client:                client,
-			Scheme:                mgr.GetScheme(),
-			DisableDefaultInstall: strings.ToLower(os.Getenv(common.DisableDefaultInstallEnvVar)) == "true",
-		}).SetupWithManager(mgr); err != nil {
-			setupLog.Error(err, "unable to create controller", "controller", "GitopsService")
-			os.Exit(1)
-		}
-	} else {
-		setupLog.Info("OpenShift Cluster not found, skipping ReconcileGitopsService controller setup")
+	if err = (&controllers.ReconcileGitopsService{
+		Client:                client,
+		Scheme:                mgr.GetScheme(),
+		DisableDefaultInstall: strings.ToLower(os.Getenv(common.DisableDefaultInstallEnvVar)) == "true",
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "GitopsService")
+		os.Exit(1)
 	}
 
 	if util.IsRouteAPIFound() {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -213,8 +213,12 @@ func main() {
 		registerComponentOrExit(mgr, monitoringv1.AddToScheme)
 	}
 
-	registerComponentOrExit(mgr, operatorsv1.AddToScheme)
-	registerComponentOrExit(mgr, operatorsv1alpha1.AddToScheme)
+	// Setup Scheme for OLM if available (verified by InspectCluster)
+	if util.IsOLMAPIFound() {
+		registerComponentOrExit(mgr, operatorsv1.AddToScheme)
+		registerComponentOrExit(mgr, operatorsv1alpha1.AddToScheme)
+	}
+
 	registerComponentOrExit(mgr, argov1alpha1api.AddToScheme)
 	registerComponentOrExit(mgr, argov1beta1api.AddToScheme)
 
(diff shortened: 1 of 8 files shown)


controllers/util/util.go
```diff
@@ -96,6 +106,40 @@ func verifyConsoleAPI() error {
 	return nil
 }
 
+func IsRouteAPIFound() bool {
+	return routeAPIFound
+}
+
+// *** THIS SHOULD ONLY BE USED FOR UNIT TESTING ***
```
svghadi: Let's move the test helper code to util_test.go

controllers/gitopsservice_controller.go
```diff
@@ -216,7 +222,12 @@ func (r *ReconcileGitopsService) Reconcile(ctx context.Context, request reconcil
 	reqLogger := logs.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling GitopsService")
 
-	// Fetch the GitopsService instance
+	// Fetch the GitopsService instance if on OpenShift cluster
+	if !argoutil.IsOpenShiftCluster() {
+		reqLogger.Info("Skip GitopsService reconcile: Not an OpenShift cluster")
+		return reconcile.Result{}, nil
+	}
```
svghadi: What do you think about not starting the gitopsservice controller itself when it's not a openshift cluster instead of skipping the reconcilation?
https://github.com/redhat-developer/gitops-operator/blob/3b6fb22628b359be0cf967f9c117498ea6e40b57/cmd/main.go#L223-L230

controllers/argocd_metrics_controller.go
```diff
@@ -83,6 +84,11 @@ func (r *ArgoCDMetricsReconciler) Reconcile(ctx context.Context, request reconci
 	reqLogger := logs.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling ArgoCD Metrics")
 
+	if !util.IsMonitoringAPIFound() {
+		reqLogger.V(1).Info("monitoring.coreos.com API not available, skipping metrics reconciliation")
+		return reconcile.Result{}, nil
+	}
+
```
svghadi: Same as previous comment

https://github.com/redhat-developer/gitops-operator/blob/3b6fb22628b359be0cf967f9c117498ea6e40b57/cmd/main.go#L240-L246

controllers/util/util.go
anandrkskd: updated, added it to test_util.go, to keep util.go clean, and test related functions in seperate file.

controllers/gitopsservice_controller.go
anandrkskd: updated

controllers/argocd_metrics_controller.go
anandrkskd: updated

cmd/main.go
```diff
@@ -133,7 +133,7 @@ func main() {
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
 	if err := util.InspectCluster(); err != nil {
-		setupLog.Info("unable to inspect cluster")
+		setupLog.Error(err, "unable to inspect cluster")
 	}
```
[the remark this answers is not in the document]
anandrkskd: This should not efect startup as this check is to inspect cluster and later use it to decide to run in openshift mode or kubernetes compatible mode.

cmd/main.go
```diff
@@ -198,18 +198,52 @@ func main() {
...
+		registerComponentOrExit(mgr, operatorsv1alpha1.AddToScheme)
+	}
+
 	registerComponentOrExit(mgr, argov1alpha1api.AddToScheme)
 	registerComponentOrExit(mgr, argov1beta1api.AddToScheme)
-	registerComponentOrExit(mgr, configv1.AddToScheme)
-	registerComponentOrExit(mgr, monitoringv1.AddToScheme)
+
+	// Setup Scheme for OpenShift Config if available
+	if util.IsOpenShiftCluster() {
```
anandf: It is better to check the API availability`config.openshift.io`,  rather than checking if it is OpenShift.  It would make the intention clear and inline with other such scheme registration.

cmd/main.go
```diff
@@ -220,30 +254,43 @@ func main() {
...
 	}
 
-	if err = (&controllers.ReconcileGitopsService{
-		Client:                client,
-		Scheme:                mgr.GetScheme(),
-		DisableDefaultInstall: strings.ToLower(os.Getenv(common.DisableDefaultInstallEnvVar)) == "true",
-	}).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "GitopsService")
-		os.Exit(1)
+	if util.IsOpenShiftCluster() {
```
anandf: This can be done in a separate PR. To conditionally reconcile `GitOpsService` CR.

anandf (changes requested):
posted few minor comments. Otherwise looks good to me.

cmd/main.go
anandrkskd: we are using config.openshift.io api availbility to check, we are naming it this way so that the switch of kubernetes and openshift is more visible to other developers.

https://github.com/redhat-developer/gitops-operator/pull/1157/changes#diff-500e1d59a2dd4d574baf6791e965cbba746d17acb6fdbfcc49e260194782195bR117

cmd/main.go
anandrkskd: understood, with this change it will work as expected in openshift cluster.
but,  we might see failures/errors when we try to run the operator in non-OCP environment if we do it in seperate PR.

controllers/util/util.go
```diff
@@ -72,21 +86,47 @@ func NewClusterVersion(version string) *configv1.ClusterVersion {
...
+	var errs []error
+	for _, check := range []func() error{
+		verifyRouteAPI,
+		verifyConsoleAPI,
+		verifyMonitoringAPI,
+		verifyTemplateAPI,
+		verifyAppsAPI,
+		verifyOAuthAPI,
+		verifyOLMAPI,
+	} {
```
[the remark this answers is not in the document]
svghadi: @anandrkskd - This looks like a valid concern. Can you take a look?

cmd/main.go
anandrkskd: updated

cmd/main.go
anandrkskd: updated

svghadi (approved):
/lgtm

anandf (approved):
LGTM

PR #1183, opened 2026-06-18
RFE-9358: Add env var to disable Swagger UI exposure

smahadik-27:
/kind feature

## Summary
- Adds `ARGOCD_SERVER_DISABLE_SWAGGER=true` environment variable to the default ArgoCD server spec
- This addresses penetration test findings (RFE-9358) where unauthenticated API schema disclosure could allow attackers to map endpoints

## Test plan
- [ ] Verify unit tests pass with `go test ./controllers/argocd/...`
- [ ] Deploy modified operator to test cluster
- [ ] Verify ArgoCD server deployment has the new env var set
- [ ] Once upstream ArgoCD supports this env var, verify /swagger-ui returns 404

## Notes
This change sets the environment variable, but requires a corresponding upstream ArgoCD change to read and act on it.

Jira: [RFE-9358](https://redhat.atlassian.net/browse/RFE-9358)

### DIFF
--- a/controllers/argocd/argocd.go
+++ b/controllers/argocd/argocd.go
@@ -162,7 +162,7 @@ func getArgoRepoServerSpec() argoapp.ArgoCDRepoSpec {
 
 func getArgoServerSpec() argoapp.ArgoCDServerSpec {
 	return argoapp.ArgoCDServerSpec{
-		Route: argoapp.ArgoCDRouteSpec{Enabled: true},
+		Route: argoapp.ArgoCDRouteSpec{Enabled: true},		// Disable Swagger UI and OpenAPI spec exposure for security hardening.		// This addresses penetration test findings (RFE-9358) where unauthenticated		// API schema disclosure could allow attackers to map endpoints and identify		// authentication bypasses.		Env: []v1.EnvVar{			{				Name:  "ARGOCD_SERVER_DISABLE_SWAGGER",				Value: "true",			},		},
 		Resources: &v1.ResourceRequirements{
 			Requests: v1.ResourceList{
 				v1.ResourceMemory: resourcev1.MustParse("128Mi"),


svghadi:
I couldn’t find any documentation regarding the `ARGOCD_SERVER_DISABLE_SWAGGER` environment variable in Argo CD. Could you point me to the docs where this behavior is defined?

smahadik-27:
I was not able to find any official Argo CD documentation that explicitly describes the behavior of the ARGOCD_SERVER_DISABLE_SWAGGER environment variable.

But, my earlier observation was based on a reproduction steps shared by customer which i performed in a test environment.

Reproduction details

1. Created an Argo CD instance in the argocd namespace.

2. created new route one that goes to the path /swagger-ui and points at a non-existent service and one with no path that points to the argocd server service.

3. Route pointing to the Argo CD server service:
**host: argocd-server-argocd.apps.testdevops10.7.india.aws.cee.support
port:
targetPort: https
tls:
insecureEdgeTerminationPolicy: Redirect
termination: reencrypt
to:
kind: Service
name: argocd-server
weight: 100
wildcardPolicy: None**

4. Route with path /swagger-ui:
**host: argocd-server-argocd.apps.testdevops10.7.india.aws.cee.support
path: /swagger-ui
port:
targetPort: https
tls:
insecureEdgeTerminationPolicy: Redirect
termination: reencrypt
to:
kind: Service
name: fake-service
weight: 100
wildcardPolicy: None**

So instead of this route-based workaround be maintained across upgrades customer wants built-in fix or configuration option in upcoming OpenShift GitOps / ArgoCD releases to disable Swagger/OpenAPI exposure?

olivergondza:
> Once upstream ArgoCD supports this env var, verify /swagger-ui returns 404

Please work with argo-cd upstream first to get this feature in. Only if/when it will be included, we can add support to the operator. Note it can be rejected by the upstream, or the variable may be renamed as a result of the review process there, hence it is better to do it later here.

PR #1194, opened 2026-06-29
Feat: RunOnInfra to use openshift.io/node-selector annotation on namespace

anandrkskd:
**What type of PR is this?**
 /kind enhancement

This PR changes `RunOnInfra` field implementation it adds `openshift.io/node-selector="node-role.kubernetes.io/infra=“` annotation on namespace and let openshift make sure the pods are scheduled on infra nodes. Previous implementation used to add node-role label to resources directly.

Fixes [GITOPS-9926](https://redhat.atlassian.net/browse/GITOPS-9926)

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test

- run `make install run`
- Update the default GitopsService CRD with `spec.runOnInfra: true`
- make sure the nodes are properly labeled
```
kubectl label node/<node-name> node-role.kubernetes.io/infra=""
```
- check if the pods are scheduled on the right Nodes.
- check if the namespace contains annotations  `openshift.io/node-selector="node-role.kubernetes.io/infra=“`

### DIFF
--- a/controllers/gitopsservice_controller.go
+++ b/controllers/gitopsservice_controller.go
@@ -249,6 +249,7 @@ func (r *ReconcileGitopsService) Reconcile(ctx context.Context, request reconcil
 	if err != nil {
 		if errors.IsNotFound(err) {
 			reqLogger.Info("Creating a new Namespace", "Name", namespace)
+			ensureInfraNodeSelectorAnnotation(namespaceRef, instance.Spec.RunOnInfra)
 			err = r.Client.Create(ctx, namespaceRef)
 			if err != nil {
 				return reconcile.Result{}, err
@@ -257,9 +258,13 @@ func (r *ReconcileGitopsService) Reconcile(ctx context.Context, request reconcil
 			return reconcile.Result{}, err
 		}
 	} else {
-		needsUpdate, updateNameSpace := ensurePodSecurityLabels(namespaceRef)
+		needsUpdate := false
+		labelUpdate, namespaceRef := ensurePodSecurityLabels(namespaceRef)
+		needsUpdate = needsUpdate || labelUpdate
+		annotationUpdate, namespaceRef := ensureInfraNodeSelectorAnnotation(namespaceRef, instance.Spec.RunOnInfra)
+		needsUpdate = needsUpdate || annotationUpdate
 		if needsUpdate {
-			err = r.Client.Update(context.TODO(), updateNameSpace)
+			err = r.Client.Update(context.TODO(), namespaceRef)
 			if err != nil {
 				return reconcile.Result{}, err
 			}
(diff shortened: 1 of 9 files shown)


controllers/gitopsservice_controller.go
```diff
@@ -1023,3 +1025,21 @@ func ensurePodSecurityLabels(namespace *corev1.Namespace) (bool, *corev1.Namespa
...
+			return true, namespace
+		}
+	} else {
+		if _, exists := namespace.Annotations[common.InfraNodeSelectorAnnotation]; exists {
+			delete(namespace.Annotations, common.InfraNodeSelectorAnnotation)
+			return true, namespace
+		}
+	}
+	return false, namespace
+}
```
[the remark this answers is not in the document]
anandrkskd: As part of the documentation it will be shared that this annotation is used, and its value will be overwritten if it was in use.

controllers/gitopsservice_controller.go
```diff
@@ -95,10 +95,15 @@ func (r *ReconcileGitopsService) SetupWithManager(mgr ctrl.Manager) error {
...
+	// check if the gitops service instance already exists, do nothing if exists, create if it doesn't
 	gitopsServiceRef := newGitopsService()
 	err := r.Client.Create(context.TODO(), gitopsServiceRef)
 	if err != nil {
-		reqLogger.Error(err, "Failed to create GitOps service instance")
+		if errors.IsAlreadyExists(err) {
+			reqLogger.Info("GitOps service instance already exists, skipping creation")
+		} else {
+			reqLogger.Error(err, "Failed to create GitOps service instance")
+		}
```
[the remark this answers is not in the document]
anandrkskd: this change only adds more defined logs for users, previous behaviour should handle it.

controllers/gitopsservice_controller.go
```diff
@@ -446,9 +457,13 @@ func (r *ReconcileGitopsService) reconcileDefaultArgoCDInstance(instance *pipeli
 			}
 		}
 
-		needsUpdate, updateNameSpace := ensurePodSecurityLabels(argocdNS)
+		needsUpdate := false
+		labelUpdate, argocdNS := ensurePodSecurityLabels(argocdNS)
+		needsUpdate = needsUpdate || labelUpdate
+		annotationUpdate, argocdNS := ensureInfraNodeSelectorAnnotation(argocdNS, instance.Spec.RunOnInfra)
+		needsUpdate = needsUpdate || annotationUpdate
```
akhilnittala: same logic as 266-270, can we make it as function ?

akhilnittala (commented):
LGTM

controllers/gitopsservice_controller.go
```diff
@@ -1013,6 +1005,15 @@ func policyRuleForBackendServiceClusterRole() []rbacv1.PolicyRule {
 	}
 }
 
+func ensureNamespaceConfig(namespace *corev1.Namespace, runOnInfra bool) bool {
```
anandf: Probably it can be named as `ensureNamespaceMetadata` as both labels and annotations that are being modified is part of `metadata`.

anandf (changes requested):
posted a minor comment otherwise looks good to me.

controllers/gitopsservice_controller.go
anandrkskd: updated, Thanks for review

PR #1206, opened 2026-07-08
feat: prevent openshift-gitops namespace creation when DISABLE_DEFAULT_ARGOCD_INSTANCE=true

Rizwana777:
**What type of PR is this?**
/kind enhancement

This PR prevents the openshift-gitops namespace from being created when DISABLE_DEFAULT_ARGOCD_INSTANCE=true is set. Previously, the flag only prevented the ArgoCD instance creation but still created the namespace.

Fixes #?
https://redhat.atlassian.net/browse/GITOPS-10435

**Test acceptance criteria**:

* [x] Unit Test
* [x] E2E Test

Fresh install with flag enabled
Set environment variable in operator Subscription
DISABLE_DEFAULT_ARGOCD_INSTANCE=true

verify
oc get namespace openshift-gitops
Expected: Error from server (NotFound)

test with make

make install
export DISABLE_DEFAULT_ARGOCD_INSTANCE=true
make run

verify
oc get namespace openshift-gitops
Expected: Error from server (NotFound)

### DIFF
--- a/controllers/gitopsservice_controller.go
+++ b/controllers/gitopsservice_controller.go
@@ -248,86 +248,114 @@ func (r *ReconcileGitopsService) Reconcile(ctx context.Context, request reconcil
 		return reconcile.Result{}, err
 	}
 
-	// Create namespace if it doesn't already exist
-	namespaceRef := newRestrictedNamespace(namespace)
-	err = r.Client.Get(ctx, types.NamespacedName{Name: namespace}, namespaceRef)
-	if err != nil {
-		if errors.IsNotFound(err) {
-			reqLogger.Info("Creating a new Namespace", "Name", namespace)
-			ensureInfraNodeSelectorAnnotation(namespaceRef, instance.Spec.RunOnInfra)
-			err = r.Client.Create(ctx, namespaceRef)
-			if err != nil {
-				return reconcile.Result{}, err
-			}
-		} else {
-			return reconcile.Result{}, err
-		}
-	} else {
-		if ensureNamespaceMetadata(namespaceRef, instance.Spec.RunOnInfra) {
-			err = r.Client.Update(context.TODO(), namespaceRef)
-			if err != nil {
-				return reconcile.Result{}, err
-			}
-		}
-	}
-
 	gitopsserviceNamespacedName := types.NamespacedName{
 		Name:      serviceName,
 		Namespace: namespace,
 	}
 
-	r.cleanKAMResources(ctx, reqLogger)
-
 	if !r.DisableDefaultInstall {
-		// Create/reconcile the default Argo CD instance, unless default install is disabled
+		// Create namespace if it doesn't already exist (only when default install is enabled)
+		namespaceRef := newRestrictedNamespace(namespace)
+		err = r.Client.Get(ctx, types.NamespacedName{Name: namespace}, namespaceRef)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				reqLogger.Info("Creating a new Namespace", "Name", namespace)
+				ensureInfraNodeSelectorAnnotation(namespaceRef, instance.Spec.RunOnInfra)
+				err = r.Client.Create(ctx, namespaceRef)
+				if err != nil {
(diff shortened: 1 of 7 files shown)


svghadi:
This will also disable console-plugin deployment, is that expected ?

Rizwana777:
> This will also disable console-plugin deployment, is that expected ?

yes, this is expected because the console plugin resources (Service, Deployment, ConfigMap) are deployed into the openshift-gitops namespace. When DISABLE_DEFAULT_ARGOCD_INSTANCE=true, the namespace does not exist and plugin resources have nowhere to be deployed, please correct me if I have understood this incorrectly.

svghadi:
> > This will also disable console-plugin deployment, is that expected ?
> 
> yes, this is expected because the console plugin resources (Service, Deployment, ConfigMap) are deployed into the openshift-gitops namespace. When DISABLE_DEFAULT_ARGOCD_INSTANCE=true, the namespace does not exist and plugin resources have nowhere to be deployed, please correct me if I have understood this incorrectly.

FYI - @anandf @keithchong

olivergondza:
@keithchong, any comments? Are you ok with such behavior?

olivergondza:
IMO, this should go in after https://github.com/redhat-developer/gitops-operator/pull/1231

keithchong:
I agree with Oliver that this should go in after https://github.com/redhat-developer/gitops-operator/pull/1231.

I am unsure why the console plugin should be disabled if the flag is set to true. In 1231, the console plugin resources are created in the operator's namespace.  So whether DISABLE_DEFAULT_ARGOCD_INSTANCE is true or false, that should not impact that?  eg. the plugin resources will still be created. 

It's just that we won't have the default Argo CD instance.  The console plugin should still exist/be enabled/work regardless.

Rizwana777:
> I agree with Oliver that this should go in after #1231.
> 
> I am unsure why the console plugin should be disabled if the flag is set to true. In 1231, the console plugin resources are created in the operator's namespace. So whether DISABLE_DEFAULT_ARGOCD_INSTANCE is true or false, that should not impact that? eg. the plugin resources will still be created.
> 
> It's just that we won't have the default Argo CD instance. The console plugin should still exist/be enabled/work regardless.

@keithchong yes after #1231 gets merged the console plugin resources will be created in the operator's namespace. I will make necessary changes in this PR once #1231 gets merged

Rizwana777:
#1231 got merged, now working on this

controllers/gitopsservice_controller.go
```diff
@@ -442,6 +459,13 @@ func (r *ReconcileGitopsService) ensureDefaultArgoCDInstanceDoesntExist() error
 		return err
 	}
 
+	// Also delete the namespace when DISABLE_DEFAULT_ARGOCD_INSTANCE is true
+	if err := r.Client.Delete(context.TODO(), argocdNS); err != nil {
+		if !errors.IsNotFound(err) {
+			return fmt.Errorf("failed to delete openshift-gitops namespace: %w", err)
+		}
+	}
+
```
jgwest, line 468: We probably don't want to delete the namespace itself when `DISABLE_DEFAULT_ARGOCD_INSTANCE is true`, there might be other stuff in there that was created outside the operator. 

We should only make sure to delete the resources that we created within the namespace, and it's up to the user to clean up any other resources in the namespace.

controllers/gitopsservice_controller.go
```diff
@@ -442,6 +459,13 @@ func (r *ReconcileGitopsService) ensureDefaultArgoCDInstanceDoesntExist() error
 		return err
 	}
 
+	// Also delete the namespace when DISABLE_DEFAULT_ARGOCD_INSTANCE is true
```
jgwest, line 462: Flagged by claude:
  ● 429 [correctness]    ensureDefaultArgoCDInstanceDoesntExist deletes the ArgoCD CR and namespace but never removes the cluster-scoped ClusterRole/ClusterRoleBinding created for the backend ServiceAccount.

jgwest (changes requested):

PR #1216, opened 2026-07-13
Mark openshift & xKS e2e tests 

anandrkskd:
**What type of PR is this?**

/kind enhancement

Operator currently only tested on OpenShift. This PR enables running e2e suite on non-OpenShift Kubernetes (xKS) by separating OpenShift-dependent tests via labels. Run --label-filter="!openshift" on xKS to skip tests requiring Routes/ConsoleLinks/SCCs/OLM. Broadens CI coverage to vanilla k8s environments.

Currently we are only labellling tests that are passing on openshift and failing on xks, this currenlty helps with the filtering.

Fixes #?

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test

NON_OLM=true ./bin/ginkgo -v --trace -p --timeout 240m -r ./test/openshift/e2e/ginkgo/sequential 
NON_OLM=true ./bin/ginkgo -v --trace -p --timeout 240m -r ./test/openshift/e2e/ginkgo/parallel

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,6 +87,9 @@ SHELL = /usr/bin/env bash -o pipefail
 # example: go list -m -f '{{.Version}}' github.com/onsi/ginkgo/v3
 GINKGO_VERSION := $(shell go list -m -f '{{.Version}}' github.com/onsi/ginkgo/v2)
 
+# XKS_LABEL_FILTER is the label filter for XKS tests.
+XKS_LABEL_FILTER ?= "!openshfit"
+
 .PHONY: all
 all: build
 
@@ -181,12 +184,12 @@ e2e-tests-parallel:
 .PHONY: e2e-xks-tests-sequential-ginkgo
 e2e-xks-tests-sequential-ginkgo: ginkgo ## Runs Ginkgo e2e sequential tests
 	@echo "Running GitOps Operator sequential Ginkgo E2E tests..."
-	$(GINKGO_CLI) -v --trace --label-filter="!notOnXKS" --timeout 240m -r ./test/openshift/e2e/ginkgo/sequential
+	$(GINKGO_CLI) -v --trace --label-filter="$(XKS_LABEL_FILTER)" --timeout 240m -r ./test/openshift/e2e/ginkgo/sequential
 
 .PHONY: e2e-xks-tests-parallel-ginkgo ## Runs Ginkgo e2e parallel tests, (Defaults to 5 runs at a time)
 e2e-xks-tests-parallel-ginkgo: ginkgo
 	@echo "Running GitOps Operator parallel Ginkgo E2E tests..."
-	$(GINKGO_CLI) -p -v -procs=5 --trace --label-filter="!notOnXKS" --timeout 60m -r ./test/openshift/e2e/ginkgo/parallel
+	$(GINKGO_CLI) -p -v -procs=5 --trace --label-filter="$(XKS_LABEL_FILTER)" --timeout 60m -r ./test/openshift/e2e/ginkgo/parallel
 
 ##@ Build
 
--- a/Makefile
+++ b/Makefile
@@ -178,6 +178,16 @@ e2e-tests-sequential:
 e2e-tests-parallel:
 	CI=prow make e2e-tests-parallel-ginkgo
 
+.PHONY: e2e-xks-tests-sequential-ginkgo
+e2e-xks-tests-sequential-ginkgo: ginkgo ## Runs Ginkgo e2e sequential tests
+	@echo "Running GitOps Operator sequential Ginkgo E2E tests..."
+	$(GINKGO_CLI) -v --trace --label-filter="!notOnXKS" --timeout 240m -r ./test/openshift/e2e/ginkgo/sequential
+
+.PHONY: e2e-xks-tests-parallel-ginkgo ## Runs Ginkgo e2e parallel tests, (Defaults to 5 runs at a time)
+e2e-xks-tests-parallel-ginkgo: ginkgo
+	@echo "Running GitOps Operator parallel Ginkgo E2E tests..."
+	$(GINKGO_CLI) -p -v -procs=5 --trace --label-filter="!notOnXKS" --timeout 60m -r ./test/openshift/e2e/ginkgo/parallel
+
 ##@ Build
 
 .PHONY: build
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,12 @@ endif
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+# GINKGO_VERSION is the version of ginkgo to use.
+# Pick ginkgo version from go.mod file.
+# Update this command when ginkgo version is updated in go.mod file. 
+# example: go list -m -f '{{.Version}}' github.com/onsi/ginkgo/v3
+GINKGO_VERSION := $(shell go list -m -f '{{.Version}}' github.com/onsi/ginkgo/v2)
+
 .PHONY: all
 all: build
 
@@ -247,7 +253,7 @@ kustomize: ## Download kustomize locally if necessary.
 GINKGO_CLI = $(shell pwd)/bin/ginkgo
 .PHONY: ginkgo
 ginkgo: ## Download ginkgo locally if necessary.
-	$(call go-get-tool,$(GINKGO_CLI),github.com/onsi/ginkgo/v2/ginkgo@v2.29.0)
+	$(call go-get-tool,$(GINKGO_CLI),github.com/onsi/ginkgo/v2/ginkgo@$(GINKGO_VERSION))
 
 
 # go-get-tool will 'go install' any package $2 and install it to $1.
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ SHELL = /usr/bin/env bash -o pipefail
 GINKGO_VERSION := $(shell go list -m -f '{{.Version}}' github.com/onsi/ginkgo/v2)
 
 # XKS_LABEL_FILTER is the label filter for XKS tests.
-XKS_LABEL_FILTER ?= "!openshfit"
+XKS_LABEL_FILTER ?= "!openshift"
 
 .PHONY: all
 all: build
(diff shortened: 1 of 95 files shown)


Makefile
```diff
@@ -172,6 +178,16 @@ e2e-tests-sequential:
 e2e-tests-parallel:
 	CI=prow make e2e-tests-parallel-ginkgo
 
+.PHONY: e2e-xks-tests-sequential-ginkgo
+e2e-xks-tests-sequential-ginkgo: ginkgo ## Runs Ginkgo e2e sequential tests
+	@echo "Running GitOps Operator sequential Ginkgo E2E tests..."
+	$(GINKGO_CLI) -v --trace --label-filter="!notOnXKS" --timeout 240m -r ./test/openshift/e2e/ginkgo/sequential
```
olivergondza: Can we rename the label `notOnXKS` to something positive in grammatical sense? `openshift`, maybe.

I envision to have multiple labels representing runtime environments or even their individual capabilities and the negation can get confusing, same as declaring what are the non-xks attributes.

anandrkskd: I agree; I was considering using the same labels, such as `openshift` for tests unique to openshift and `xKS` or any other platform-dependent labels for their particular environment. 
I talked this over with @varshab1210, and she suggested that we create a label indicating where not to run it, like `notOnXKS`, `notOnOpenshift` or something along those lines.

varshab1210: > I envision to have multiple labels representing runtime environments or even their individual capabilities and the negation can get confusing, same as declaring what are the non-xks attributes.

I made this suggestion to reduce label usage. If we consider adding `openshift` and `xKS` labels we would need to add them on every test going forward because we expect features to work on both platforms and hence label it accordingly  (unless I misunderstood the intention), whereas if we use something like `skip_on_xKS`, it needs to be added only when a particular feature is not going to be supported on xKS or `skip_on_openshift` for that matter for other way around. 
@anandrkskd when I initially reviewed the PR, I thought you were using `openshift` as a way to skip the run on `xKS`. That seemed a little confusing to me. 
I am okay either way, as long as we have a README.md describing the scenarios for labeling

anandrkskd: Yes, currently we are using it to skip test that are currently only compatible with openshift, we should only add labels for tests that are platform specific, either openshift or xks. Labels are not required on every test aa most test should be generic and should work on both the platforms. 
I will update the Readme.md to make it more transparent as an action item.

olivergondza: We seem to have a different understanding about the semantics of such labels.

What does `openshift` mean, for example? Is the test _specific_ for openshift so it does not need to run on any other platforms, or it means the test is _relevant_ for openshift and potentially for other platforms as well.

I find the second meaning less practical because - as Varsha pointed out - most of the tests would have most of the labels and the lists will only ever expand. Also, forgetting to add label(s) will cause tests are not executed creating a false sense of security.

I suggest the meaning of _specfic_: "this tests depends on a feature specific to given environment". That way, most of the tests depending on k8s essentials can remain unlabeled. And we will only bother the authors of platform-specific tests. The downside here is that the test executions will have to exclude `N-1` platforms when launching, like `--label-filter=!openshift && !...`, but like not a big problem given this is done on handful of places (one?) and for handful of non-k8s environments. Here, forgetting to add label to test (forgetting to indicate that tests require special features of a platform) will cause test failure on other platforms, pointing out the mistake.

anandrkskd: > I suggest the meaning of specfic: "this tests depends on a feature specific to given environment". That way, most of the tests depending on k8s essentials can remain unlabeled. And we will only bother the authors of platform-specific tests. The downside here is that the test executions will have to exclude N-1 platforms when launching, like --label-filter=!openshift && !..., but like not a big problem given this is done on handful of places (one?) and for handful of non-k8s environments. Here, forgetting to add label to test (forgetting to indicate that tests require special features of a platform) will cause test failure on other platforms, pointing out the mistake.

I agree with what you have mentioned above, labeling depends on a feature specific to platform. When we label a test with openshift it should not run on other platforms, or else it will fails.

varshab1210: My only concern is people adding the label `openshift` thinking "Yes, the feature is supported on Openshift" instead of thinking if the feature is _only_ supported on Openshift. We don't want to skip a test run on xKS accidentally. Or someone going through the test thinking tests labelled as `openshift` are the ones supported on Openshift. A README.md would help clarify this though so its fine by me.

olivergondza: Agreed this needs to be documented thoroughly. @anandrkskd, could you?

> thinking "Yes, the feature is supported on Openshift"

If we cant to be super clear about this, we can introduce a dummy wrapper function for `Label` like `Requires("openshift")`. But I am not sure it is worth it.

anandrkskd: I will update the readme, if it is still not clear, we can work towards a wrapper function as last resort.
WDYT?

test/openshift/e2e/ginkgo/sequential/1-120_repo_server_system_ca_trust.go
```diff
@@ -81,7 +81,7 @@ var (
 )
 
 var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
-	Context("1-120_repo_server_system_ca_trust", Label("notOnXKS"), func() {
+	Context("1-120_repo_server_system_ca_trust", Label("openshfit"), func() {
```
olivergondza: I am checking the label assignment in `1-120_repo_server_system_ca_trust.go` only, and it is not correct. None of the scenarios in the file is openshift specific.

Also, what is the point to label both `Context` and the `It` inside it with the same label? I would expect that `--label-filter=!openshift` will skip all `It`s in the `Context("...", Label("openshfit"), func() {`, no?

olivergondza (changes requested):

test/openshift/e2e/ginkgo/sequential/1-120_repo_server_system_ca_trust.go
anandrkskd: >Also, what is the point to label both Context and the It inside it with the same label? I would expect that --label-filter=!openshift will skip all Its in the Context("...", Label("openshfit"), func() {, no?

yes, it slipped from my changes. If the context if labeled we dont need to label tests sperately.

>I am checking the label assignment in 1-120_repo_server_system_ca_trust.go only, and it is not correct. None of the scenarios in the file is openshift specific.

These changes are in its initial stage, we want to start testing on PRs ASAP.
For sequential we are labeling the tests that are failing on xKS runs. We will be going through each test to port or make changes to make them platform independent.
I am considering moving the sequential to another PR, but in that case we will skip testing the senarios that are working.
cc: @varshab1210

anandrkskd: Also the test that you mentioned is not openshift spicific but uses csv, that we are not using for xKS
 
```
image, imageVersion = getArgoCDComponentImageAndVersion()
```
And thats the reason we labeled it as openshift.
We need to itirate over the existing tests multiple times, and make sure all tests are ported or are labeled correctly going forward.

olivergondza: > For sequential we are labeling the tests that are failing on xKS runs.

Gotcha. I would like to see a `// TODO` marker there to clarify which of the "openshift" tests are for later review.

In case of these tests, it can be the `ClusterTrustBundle` feature flag that is causing the failures.

olivergondza: > Also the test that you mentioned is not openshift spicific but uses csv, that we are not using for xKS

Fair enough. Might require blending with the one in argocd-operator, that does not.

anandrkskd: I have updated the files where i felt we should update the test to make compatible for xks run or to verify.
 
Can give this PR another look.

olivergondza (commented):
Removing the request for changes. Thanks!

test/openshift/e2e/ginkgo/fixture/fixture.go
```diff
@@ -988,6 +993,22 @@ func RunningOnOpenShift() bool {
...
+	crdList := crdv1.CustomResourceDefinitionList{}
+	Expect(k8sClient.List(context.Background(), &crdList)).To(Succeed())
+
+	olmAPIsFound := 0
+	for _, crd := range crdList.Items {
+		if strings.Contains(crd.Spec.Group, "operators.coreos.com") {
+			olmAPIsFound++
+		}
+	}
+	return olmAPIsFound > 0
```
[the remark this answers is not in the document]
svghadi: @anandrkskd - this looks like a valid suggestion. Also I don't see `IsOperatorRunningOnOLM` used anywhere in the code. Is this dead code?

test/openshift/e2e/ginkgo/parallel/1-063_validate_dex_liveness_probe_test.go
```diff
@@ -30,8 +30,8 @@ import (
 )
 
 var _ = Describe("GitOps Operator Parallel E2E Tests", func() {
-
-	Context("1-063_validate_dex_liveness_probe_test", func() {
+	// TODO: make it XKS compatible
+	Context("1-063_validate_dex_liveness_probe_test", Label("openshift"), Label("dex"), func() {
```
svghadi: Any specific reason for dex label?

svghadi (commented):
Overall looks good to me. Left some comments. 

Also, will make these tests xks compatible?

test/openshift/e2e/ginkgo/parallel/1-063_validate_dex_liveness_probe_test.go
anandrkskd: no specific reson, it was part of testing, & can be removed.

test/openshift/e2e/ginkgo/fixture/fixture.go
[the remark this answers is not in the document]
anandrkskd: We are currently setting NON_OLM env var in our test command to let the tests know we are not running non-olm installation method(`make deploy` or `helm chart`), going forward we can use this method to detect operators installation method and remove reduce dependency on the env var.
As an action Item I will add a TODO to make update this and use it in upcoming changes in test suits.

anandrkskd:
>Also, will make these tests xks compatible?

Yes we will spend more time on testing and increase the tests coverage in comming up PRs.

svghadi (dismissed):
/lgtm

PR #1217, opened 2026-07-14
CI: Github action job tests for xks on push to master and release branch

anandrkskd:
**What type of PR is this?**

/kind enhancement

This PR adds CI Jobs for push to master and release branch, We are going to use existing kind automation for cluster for testing with 3 nodes.
- adds a script for eks cluster creation & deletion 

Fixes #?

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test


hack/scripts/eks-cluster.sh
```diff
@@ -0,0 +1,96 @@
...
+
+cmd_delete() {
+  echo "==> This will DELETE cluster '${CLUSTER_NAME}' in ${REGION} and all associated resources"
+  if [[ "${FORCE:-false}" != "true" ]]; then
+    read -rp "    Continue? [y/N] " confirm
+    [[ "${confirm}" =~ ^[Yy]$ ]] || { echo "Aborted."; exit 0; }
+  fi
+
+  echo "==> Deleting EKS cluster: ${CLUSTER_NAME}"
+  eksctl delete cluster \
```
AdamSaleh, line 91: What happens if this fails? Should we have some sort of retry-logic here?

hack/scripts/eks-cluster.sh
```diff
@@ -0,0 +1,96 @@
...
+    echo "ERROR: ${cmd} not found in PATH" >&2
+    exit 1
+  fi
+done
+
+cmd_create() {
+  echo "==> Creating EKS cluster: ${CLUSTER_NAME} in ${REGION}"
+  echo "    Kubernetes ${K8S_VERSION}, ${NODE_COUNT}x ${NODE_TYPE}, ${NODE_VOLUME_SIZE}GiB volumes"
+
+  eksctl create cluster \
```
AdamSaleh: Should we have some timeout logic here? Also cleanup on timeout/faillure?

.github/workflows/xks-push-ci-automation.yaml
```diff
@@ -0,0 +1,215 @@
+name: Build, Deploy and Test on EKS
+
+on:
+  push:
+    branches:
+      - "master"
+      - "v*"
+
```
AdamSaleh: I think we should use some sort of concurrency limitation, so that there is limit on the number of clusters we could create i.e. https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-using-concurrency-and-the-default-behavior

hack/scripts/eks-cluster.sh
anandrkskd: added, on timeout the cluster will be cleaned up, as this is not part of the CI, we can re run the script to create cluster.

.github/workflows/xks-push-ci-automation.yaml
anandrkskd: we are not using this eks for CI so this is not requried.

hack/scripts/eks-cluster.sh
anandrkskd: added timeout logic

svghadi (approved):

PR #1218, opened 2026-07-15
feat(redis-ha-proxy): Configure Redis Ha Proxy TLS settings

akhilnittala:
**What type of PR is this?**
/kind enhancement

Configure TLS settings for redis ha proxy.

Fixes #?
https://redhat.atlassian.net/browse/GITOPS-10289
**Test acceptance criteria**:

* [x] Unit Test
* [ ] E2E Test

> deploy redis ha proxy pods with ha.enabled to true in argocd CR.


build/redis/haproxy.cfg.tpl
```diff
@@ -1,6 +1,27 @@
...
+    ssl-default-server-ciphersuites {{.tlsCiphers}}
+{{- else}}
+    # TLS 1.2 and below cipher lists
+    ssl-default-bind-ciphers {{.tlsCiphers}}
+    ssl-default-server-ciphers {{.tlsCiphers}}
+
+    # Also configure TLS 1.3 cipher suites when TLS 1.3 is negotiated
+    ssl-default-bind-ciphersuites {{.tlsCiphers}}
+    ssl-default-server-ciphersuites {{.tlsCiphers}}
+{{- end}}
```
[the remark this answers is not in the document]
akhilnittala: we are passing min version of tls only that means when 1.2 is enabled it can also compatible with 1.3 as well, so adding wont cause any harm

build/redis/haproxy.cfg.tpl
```diff
@@ -1,6 +1,27 @@
...
+{{- if eq .tlsMinVersion "1.3"}}
+    # TLS 1.3 cipher suites
+    ssl-default-bind-ciphersuites {{.tlsCiphers}}
+    ssl-default-server-ciphersuites {{.tlsCiphers}}
+{{- else}}
+    # TLS 1.2 and below cipher lists
+    ssl-default-bind-ciphers {{.tlsCiphers}}
+    ssl-default-server-ciphers {{.tlsCiphers}}
+
+    # Also configure TLS 1.3 cipher suites when TLS 1.3 is negotiated
```
anandf, line 20: Why TLS 1.3 ciphers are also included by default ?

anandf (changes requested):

build/redis/haproxy.cfg.tpl
akhilnittala: when minversion is set to 1.2 means minversion is 1.2 and maxversion as it is unset can be compatible for 1.3 as well,
something like if tls1.2 is minversion it is compatible for 1.2 and 1.3, but if tls1.3 is minversion then only compatible for 1.3


If ssl-max-ver is not set, HAProxy imposes no upper limit on the TLS version. It will negotiate the highest TLS version supported by both:

akhilnittala:
this pr pulled latest CRD sturctures from config/crd/bases directory of argocd-operator, so some changes related to CRD's are related to those

anandf (approved):
LGTM

PR #1224, opened 2026-07-18
feat(gitops-console-plugin): Configure SSL protocol and ciphersuites for plugin component

akhilnittala:
**What type of PR is this?**
/kind enhancement

Configure SSL Protocol and cipher suites based on apiserver tls minversion and ciphersuites

Fixes #?
https://redhat.atlassian.net/browse/GITOPS-10473
**Test acceptance criteria**:

* [x] Unit Test
* [ ] E2E Test

deploy gitops operator and verify the ssl connection with apiserver tls minversion and ciphers


controllers/consoleplugin.go
```diff
@@ -396,6 +434,7 @@ func (r *ReconcileGitopsService) reconcileDeployment(cr *pipelinesv1alpha1.Gitop
 			existingPluginDeployment.Spec.Replicas = newPluginDeployment.Spec.Replicas
 			existingPluginDeployment.Spec.Selector = newPluginDeployment.Spec.Selector
 			existingSpecTemplate.Labels = newSpecTemplate.Labels
+			existingSpecTemplate.ObjectMeta.Annotations["httpd-cfg-hash"] = newSpecTemplate.ObjectMeta.Annotations["httpd-cfg-hash"]
```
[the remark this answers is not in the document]
akhilnittala: already doing it in 394 line 	if newPluginDeployment.Spec.Template.ObjectMeta.Annotations == nil {
		newPluginDeployment.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
	}

controllers/consoleplugin.go
```diff
@@ -487,9 +526,18 @@ func (r *ReconcileGitopsService) reconcileConsolePlugin(instance *pipelinesv1alp
...
+	hash := md5.New()
+	for key, value := range cm.Data {
+		io.WriteString(hash, key)
+		io.WriteString(hash, value)
+	}
+	return fmt.Sprintf("%x", hash.Sum(nil))
+}
+
+func (r *ReconcileGitopsService) reconcileConfigMap(instance *pipelinesv1alpha1.GitopsService, request reconcile.Request, newPluginConfigMap *corev1.ConfigMap) (reconcile.Result, error) {
 	reqLogger := logs.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
```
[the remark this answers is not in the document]
akhilnittala: will do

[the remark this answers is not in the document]
akhilnittala: modified

controllers/consoleplugin.go
```diff
@@ -529,15 +597,18 @@ func (r *ReconcileGitopsService) reconcilePlugin(instance *pipelinesv1alpha1.Git
...
 	}
 
-	if result, err := r.reconcileDeployment(instance, request); err != nil {
+	if result, err := r.reconcileDeployment(instance, request, newPluginConfigMap); err != nil {
 		return result, err
 	}
 
-	if result, err := r.reconcileConfigMap(instance, request); err != nil {
+	if result, err := r.reconcileConfigMap(instance, request, newPluginConfigMap); err != nil {
 		return result, err
```
aali309: should we rearrange this to reconcile ConfigMap before Deployment? 
```suggestion
	if result, err := r.reconcileConfigMap(instance, request, newPluginConfigMap); err != nil {
         return results, err

	if result, err := r.reconcileDeployment(instance, request, newPluginConfigMap); err != nil {
        return result, err
}
```

aali309 (commented):
Thanks @akhilnittala, the approach looks right, left a comment

controllers/consoleplugin.go
akhilnittala: done @aali309

controllers/consoleplugin.go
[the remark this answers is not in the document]
aali309: @akhilnittala, what do you think about addressing this?

controllers/consoleplugin.go
```diff
@@ -240,18 +244,52 @@ func securityContextForPlugin() *corev1.SecurityContext {
...
+		httpdConfigBase += "\n\tSSLProtocol -all +TLSv1 +TLSv1.1 +TLSv1.2 +TLSv1.3"
+	case "VersionTLS11":
+		httpdConfigBase += "\n\tSSLProtocol -all +TLSv1.1 +TLSv1.2 +TLSv1.3"
+	case "VersionTLS12":
+		httpdConfigBase += "\n\tSSLProtocol -all +TLSv1.2 +TLSv1.3"
+	case "VersionTLS13":
+		httpdConfigBase += "\n\tSSLProtocol -all +TLSv1.3"
+	}
+	if TLSVersionToPlugin(r.TLSMinVersion) != "TLSv1.3" && strings.Join(r.TLSCiphers, ":") != "" {
+		httpdConfigBase += fmt.Sprintf("\n\tSSLCipherSuite %s", strings.Join(r.TLSCiphers, ":"))
```
aali309: Just a question here to confirm if `Intermediate/Modern` profile ciphers work with httpd as-is, or do we need to filter/map them?

akhilnittala: i tested with as-is, no pod crashed, for go cipher list may be we need to map them to go based ciphers.

aali309:
Thanks @akhilnittala, LGTM! Extra eyes will be  very helpful.

cmd/main.go
```diff
@@ -312,6 +312,8 @@ func main() {
 			Client:                client,
 			Scheme:                mgr.GetScheme(),
 			DisableDefaultInstall: strings.ToLower(os.Getenv(common.DisableDefaultInstallEnvVar)) == "true",
```
svghadi, line 314: same suggestion as from backend pr, can we use the profile object itsel? Also we might have to handle case where tls settings is disabled at cluster level

akhilnittala: will be handled.

controllers/consoleplugin.go
```diff
@@ -359,6 +398,13 @@ func (r *ReconcileGitopsService) reconcileDeployment(cr *pipelinesv1alpha1.Gitop
 		newPluginDeployment.Spec.Template.Spec.Tolerations = cr.Spec.Tolerations
 	}
 
+	// ADD THIS: Get ConfigMap and add hash to pod template annotations
```
svghadi, line 388: I think we can avoid all this complexity by computing the hash directly in the ConfigMap reconciliation function instead of storing it. During each reconciliation, we can compare the hash of the existing and desired  httpd configs or simply compare the strings, if possible. That should be sufficient to detect drift and update the ConfigMap.

controllers/consoleplugin.go
```diff
@@ -240,18 +244,53 @@ func securityContextForPlugin() *corev1.SecurityContext {
...
+	case "VersionTLS10":
+		httpdConfigBase += "\n\tSSLProtocol -all +TLSv1 +TLSv1.1 +TLSv1.2 +TLSv1.3"
+	case "VersionTLS11":
+		httpdConfigBase += "\n\tSSLProtocol -all +TLSv1.1 +TLSv1.2 +TLSv1.3"
+	case "VersionTLS12":
+		httpdConfigBase += "\n\tSSLProtocol -all +TLSv1.2 +TLSv1.3"
+	case "VersionTLS13":
+		httpdConfigBase += "\n\tSSLProtocol -all +TLSv1.3"
+	}
+	if TLSVersionToPlugin(minVersionTLS) != "TLSv1.3" && strings.Join(r.CentralTLSProfile.Ciphers, ":") != "" {
```
svghadi: Do we require a conversion? Can't we directly us the actual values to compare?

controllers/consoleplugin.go
akhilnittala: The hash annotation isn't only used to detect ConfigMap drift; it's used to trigger a Deployment rollout when the ConfigMap content changes. Since httpd reads its configuration at startup, updating only the ConfigMap wouldn't restart the pods. Unless we introduce another restart mechanism, we still need a pod template change (such as the hash annotation) to ensure the updated configuration is applied.

controllers/consoleplugin.go
akhilnittala: made changes, removed conversion.

anandf (approved):
LGTM

aali309 (approved):

keithchong (approved):
Thanks @akhilnittala and @aali309.  LGTM.

svghadi (approved):
/lgtm
/approve

PR #1231, opened 2026-07-22
feat: decouple console plugin from default ArgoCD instance 

Rizwana777:
**What type of PR is this?**
/kind enhancement

Moves console plugin resources (Deployment, Service, ConfigMap) from 'openshift-gitops' namespace to the operator's own namespace (openshift-gitops-operator) so the plugin operates independently of the default ArgoCD instance. This is a prerequisite for allowing the default ArgoCD instance to be disabled without breaking the console plugin.

- Made plugin functions support namespaces.
- Added 'PluginNamespace' field to the reconciler, resolved from the operator's serviceaccount namespace
- Added cleanup of old plugin resources from 'openshift-gitops'
- Ensures plugin namespace exists before creating resources
- ConsolePlugin CR backend now points to the operator namespace

Fixes #?

JIRA -https://redhat.atlassian.net/browse/GITOPS-9846

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test

make install 
make run

-  kubectl get deployment gitops-plugin -n openshift-gitops-operator
   NAME            READY   UP-TO-DATE   AVAILABLE   AGE
   gitops-plugin   1/1     1            1           19m

-  kubectl get service gitops-plugin -n openshift-gitops-operator 
   NAME            TYPE         CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
  gitops-plugin   ClusterIP   172.xyz.83   <none>        9001/TCP         19m

-  kubectl get configmap httpd-cfg -n openshift-gitops-operator
    NAME        DATA   AGE
    httpd-cfg   1      20m

- kubectl get consoleplugin gitops-plugin -o jsonpath='{.spec.backend.service.namespace}'
   openshift-gitops-operator

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -311,12 +311,26 @@ func main() {
 		}
 	}
 
+	pluginNamespace := "openshift-gitops-operator"
+	data, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+	if err != nil {
+		if os.IsNotExist(err) {
+			setupLog.Info(fmt.Sprintf("Unable to retrieve the operator's running namespace via serviceaccount: %v. Using default namespace '%s'. This is expected when running locally.", err, pluginNamespace))
+		} else {
+			setupLog.Error(err, "Error retrieving operator's running namespace")
+			os.Exit(1)
+		}
+	} else {
+		pluginNamespace = strings.TrimSpace(string(data))
+	}
+
 	if util.IsOpenShiftCluster() {
 		if err = (&controllers.ReconcileGitopsService{
 			Client:                client,
 			Scheme:                mgr.GetScheme(),
 			DisableDefaultInstall: strings.ToLower(os.Getenv(common.DisableDefaultInstallEnvVar)) == "true",
 			CentralTLSProfile:     profile,
+			PluginNamespace:       pluginNamespace,
 		}).SetupWithManager(mgr); err != nil {
 			setupLog.Error(err, "unable to create controller", "controller", "GitopsService")
 			os.Exit(1)
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -311,17 +311,10 @@ func main() {
 		}
 	}
 
-	pluginNamespace := "openshift-gitops-operator"
-	data, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+	pluginNamespace, err := util.GetOperatorNamespace()
 	if err != nil {
-		if os.IsNotExist(err) {
-			setupLog.Info(fmt.Sprintf("Unable to retrieve the operator's running namespace via serviceaccount: %v. Using default namespace '%s'. This is expected when running locally.", err, pluginNamespace))
-		} else {
-			setupLog.Error(err, "Error retrieving operator's running namespace")
-			os.Exit(1)
-		}
-	} else {
-		pluginNamespace = strings.TrimSpace(string(data))
+		setupLog.Error(err, "Error retrieving operator's running namespace")
+		os.Exit(1)
 	}
 
 	if util.IsOpenShiftCluster() {
--- a/controllers/consoleplugin.go
+++ b/controllers/consoleplugin.go
@@ -552,7 +552,6 @@ func getConfigMapHash(cm *corev1.ConfigMap) string {
 
 func (r *ReconcileGitopsService) reconcileConfigMap(instance *pipelinesv1alpha1.GitopsService, request reconcile.Request, newPluginConfigMap *corev1.ConfigMap) (reconcile.Result, error) {
 	reqLogger := logs.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
-	newPluginConfigMap = r.pluginConfigMap(r.PluginNamespace)
 
 	if err := controllerutil.SetControllerReference(instance, newPluginConfigMap, r.Scheme); err != nil {
 		return reconcile.Result{}, err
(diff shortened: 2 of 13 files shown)


controllers/consoleplugin.go
```diff
@@ -521,6 +521,39 @@ func (r *ReconcileGitopsService) reconcileConfigMap(instance *pipelinesv1alpha1.
...
+		return
+	}
+
+	reqLogger := logs.WithValues()
+
+	oldDeploy := &appsv1.Deployment{}
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: gitopsPluginName, Namespace: serviceNamespace}, oldDeploy); err == nil {
+		reqLogger.Info("Cleaning up old plugin Deployment from previous namespace", "Namespace", serviceNamespace)
+		if err := r.Client.Delete(ctx, oldDeploy); err != nil && !errors.IsNotFound(err) {
+			reqLogger.Error(err, "Failed to delete old plugin Deployment", "Namespace", serviceNamespace)
```
jgwest: Should return error, rather than just logging it

controllers/consoleplugin.go
```diff
@@ -521,6 +521,39 @@ func (r *ReconcileGitopsService) reconcileConfigMap(instance *pipelinesv1alpha1.
...
+		}
+	}
+
+	oldSvc := &corev1.Service{}
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: gitopsPluginName, Namespace: serviceNamespace}, oldSvc); err == nil {
+		reqLogger.Info("Cleaning up old plugin Service from previous namespace", "Namespace", serviceNamespace)
+		if err := r.Client.Delete(ctx, oldSvc); err != nil && !errors.IsNotFound(err) {
+			reqLogger.Error(err, "Failed to delete old plugin Service", "Namespace", serviceNamespace)
```
jgwest: Should return error, rather than just logging it

controllers/consoleplugin.go
```diff
@@ -521,6 +521,39 @@ func (r *ReconcileGitopsService) reconcileConfigMap(instance *pipelinesv1alpha1.
...
+		}
+	}
+
+	oldCM := &corev1.ConfigMap{}
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: httpdConfigMapName, Namespace: serviceNamespace}, oldCM); err == nil {
+		reqLogger.Info("Cleaning up old plugin ConfigMap from previous namespace", "Namespace", serviceNamespace)
+		if err := r.Client.Delete(ctx, oldCM); err != nil && !errors.IsNotFound(err) {
+			reqLogger.Error(err, "Failed to delete old plugin ConfigMap", "Namespace", serviceNamespace)
```
jgwest: Should return error, rather than just logging it

controllers/gitopsservice_controller.go
```diff
@@ -326,9 +349,16 @@ func (r *ReconcileGitopsService) Reconcile(ctx context.Context, request reconcil
...
-	} else {
-		return r.reconcilePlugin(instance, request)
 	}
+
+	result, err := r.reconcilePlugin(instance, request)
+	if err != nil {
+		return result, err
+	}
+
+	r.cleanupOldPluginResources(ctx)
```
jgwest: Can return error here if present

test/openshift/e2e/ginkgo/sequential/1-124_validate_console_plugin_in_operator_namespace_test.go
```diff
@@ -0,0 +1,81 @@
...
+			oldDepl := &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{Name: "gitops-plugin", Namespace: "openshift-gitops"}}
+			Eventually(oldDepl).Should(k8sFixture.NotExistByName())
+
+			oldSvc := &corev1.Service{ObjectMeta: metav1.ObjectMeta{Name: "gitops-plugin", Namespace: "openshift-gitops"}}
+			Eventually(oldSvc).Should(k8sFixture.NotExistByName())
+
+			oldCM := &corev1.ConfigMap{ObjectMeta: metav1.ObjectMeta{Name: "httpd-cfg", Namespace: "openshift-gitops"}}
+			Eventually(oldCM).Should(k8sFixture.NotExistByName())
+		})
+	})
```
jgwest, line 179: Can you add a second test that creates plugin resources in the old namespace (simulating an old installation) and verifies they are cleaned up?

jgwest (changes requested):

controllers/consoleplugin.go
```diff
@@ -521,6 +521,39 @@ func (r *ReconcileGitopsService) reconcileConfigMap(instance *pipelinesv1alpha1.
 	return reconcile.Result{}, nil
 }
 
+// cleanupOldPluginResources removes plugin resources from the old namespace (openshift-gitops) after they have been moved to the operator namespace, since owner references on the cluster-scoped GitopsService CR won't trigger garbage collection.
+func (r *ReconcileGitopsService) cleanupOldPluginResources(ctx context.Context) {
```
anandrkskd: nit: It seems RN `cleanupOldPluginResources` func will be called on every reconciliation.  can we add a check to return early if the cleanup was done successfully?

new unexported field `pluginCleanupComplete` in ReconcileGitopsService, that is set to true when the cleanup completes successfully.

cmd/main.go
```diff
@@ -307,11 +307,25 @@ func main() {
...
+		if os.IsNotExist(err) {
+			setupLog.Info(fmt.Sprintf("Unable to retrieve the operator's running namespace via serviceaccount: %v. Using default namespace '%s'. This is expected when running locally.", err, pluginNamespace))
+		} else {
+			setupLog.Error(err, "Error retrieving operator's running namespace")
+			os.Exit(1)
+		}
+	} else {
+		pluginNamespace = strings.TrimSpace(string(data))
+	}
+
```
anandrkskd: can we move this to utils.go, we have used this same logic in multiple places, suite_test.go in e2e and nondefaulte2e dirs.

anandrkskd (approved):
some nit comments, changes LGTM

controllers/consoleplugin.go
Rizwana777: The Get calls in cleanupOldPluginResources go through controller-runtime split client, which reads from the informer cache rather than making direct API server calls. As a result, the per-reconcile overhead is negligible since these are in-memory lookups. Introducing a pluginCleanupComplete boolean flag would cause it to be set to true during the first reconcile if no old resources exist. From that point on, cleanup would be permanently skipped for the lifetime of the operator pod. This means that if old resources appear later (for example, due to an upgrade race or manual re-creation), they would never be cleaned up until the operator pod is restarted.

Given the negligible cost of cache reads and the risk of missing legitimate cleanup, I think it's safer to continue checking on every reconcile.

Let me know what you think about this

anandrkskd: Correct me if I am wrong, only configMap and secrets are cached when memoryOptimization is enabled, other resources still goes to cluster when a GET call is made.

Rizwana777:
ci is failing with flaky e2e tests

jgwest:
```
go: k8s.io/kubernetes@v1.36.1: read "https://proxy.golang.org/k8s.io/kubernetes/@v/v1.36.1.mod": stream error: stream ID 393; INTERNAL_ERROR; received from peer```

/retest

**Comment by jgwest** [comment]
@Rizwana777 can you investigate this failure in E2E test:
```
  [38;5;243m/go/src/github.com/redhat-developer/gitops-operator/test/openshift/e2e/ginkgo/sequential/1-124_validate_console_plugin_in_operator_namespace_test.go:56[0m
  [38;5;9m[FAIL][0m [0mGitOps Operator Sequential E2E Tests [38;5;243m1-124_validate_console_plugin_in_operator_namespace [38;5;9m[1m[It] cleans up plugin resources from the old namespace when they exist (simulating upgrade from older version)[0m
  [38;5;243m/go/src/github.com/redhat-developer/gitops-operator/test/openshift/e2e/ginkgo/sequential/1-124_validate_console_plugin_in_operator_namespace_test.go:99[0m
```

jgwest (approved):
LGTM, thanks @Rizwana777!

PR #1239, opened 2026-07-28
feat: Migrate from manifest v0 to v1 format

olivergondza:
**What type of PR is this?**

> /kind bug
> /kind cleanup
> /kind failing-test
/kind enhancement
> /kind documentation
> /kind code-refactoring

Avoid using the OLMv1 and move towards OLMv2.

Co-authored-by: Cursor

https://redhat.atlassian.net/browse/GITOPS-10517
https://redhat.atlassian.net/browse/GITOPS-10516

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test


olivergondza:
@svghadi, @jgwest, can you please take a look?

This is a first step in the above Epic, addressing what I consider are the most pressing aspects before the OLMv0 deprecation.

bundle/bundle.Dockerfile
```diff
@@ -1,24 +1,24 @@
+# TODO: Remove this and use the bundle.Dockerfile in the root of the project.
+# To be dropped after the openshift-ci have started reding the other one.
+
 FROM scratch
 
+# Core bundle labels.
 LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
 LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
-LABEL operators.operatorframework.io.bundle.package.v1=openshift-gitops-operator
```
svghadi, line 6: We should preserve this name. In downstream we use `openshift-gitops-operator`
https://github.com/rh-gitops-midstream/release/blob/1e290203760532335b3d5bfcfc3919ee92a678ad/containers/gitops-operator-bundle/Dockerfile#L7

Containerfile.plugin
svghadi, line 1: We can actually get rid of this file. It was added for konflux builds which don't run now in this repo.

svghadi (commented):
I have a question regarding backward compatibility. Can the new bundle and catalog images containing OLM v1 changes be used to install the operator on OLM v0 clusters, or do we need to publish separate images for OLM v0 and OLM v1?

olivergondza:
> I have a question regarding backward compatibility. Can the new bundle and catalog images containing OLM v1 changes be used to install the operator on OLM v0 clusters, or do we need to publish separate images for OLM v0 and OLM v1?

As discussed offline, OCP 4.18 is the first release where OLMv1 is GA. So there will be problems installing this on older ones, where OLMv1 is a Tech Preview. However, latest gitops-operators depends on 4.18+, so there is no need to support both OLM versions, provided we do not backport this change.

bundle/bundle.Dockerfile
olivergondza: Based on your comment, I tried to get it aligned with production build, but the prefix makes the computed name+version too long for the 33-ish character limit, that was surprisingly hard to workaround.

The `make olm-deploy` and the manual procedure we had before, was using the other "bundle.Dockerfile" with `gitops-operator` as a package name. So this does not change how the developers deploy the operator.  I suggest keeping this as is, because a) it worked this way for years, and b) it is not used for other purposed than temporary deploying during development.

olivergondza:
@svghadi, can you please take a look?

olivergondza:
@varshab1210, PTAL for QE perspective. IMO, it should not influence you...

bundle/bundle.Dockerfile
```diff
@@ -1,24 +1,24 @@
...
-# delivered via an index image
-LABEL com.redhat.delivery.operator.bundle=true
-# This second label tells the pipeline which versions of OpenShift the operator supports.
-# This is used to control which index images should include this operator.
-LABEL com.redhat.openshift.versions="v4.8"
-# This third label tells the pipeline that this operator should *also* be supported on OCP 4.4 and
-# earlier.  It is used to control whether or not the pipeline should attempt to automatically
-# backport this content into the old appregistry format and upload it to the quay.io application
-# registry endpoints.
-LABEL com.redhat.delivery.backport=true
```
chengfang, line 24: is the safe to remove the 3 labels? The last 2 seems obsolete and may be ok, but the first one looks still important, from its comment above.

A quick search says "Yes, it is required if you are publishing, distributing, or certifying your Kubernetes Operator bundle through the Red Hat OpenShift ecosystem."

olivergondza: I believe it is safe. As it turned out, this file is only used by openshift-ci - it is not even used for development, not for releasing. See the TODO comment at the top, once openshift-ci is pointed to the other file this one will be unused.

PR #1278, opened 2026-09-06
fix: omit nonResourceURLs from namespaced Roles

alkakumari016:
**What type of PR is this?**
 /kind bug

The application-controller Role copies rules from the aggregated admin ClusterRole. Third-party ClusterRoles labeled aggregate-to-admin can contribute nonResourceURLs, which are illegal on a Role, so the API server rejects it and Argo CD reconcile hard-fails (including Dex / OpenShift OAuth).

This PR fixes that by stripping nonResourceURLs from the namespace role while copying from admin ClusterRole.

Fixes #?
https://github.com/argoproj-labs/argocd-operator/issues/2269
https://redhat.atlassian.net/browse/GITOPS-10618
**Test acceptance criteria**:

* [x] Unit Test
* [x] E2E Test

Steps to Reproduce

Install the affected version of the OpenShift GitOps Operator 1.19.3.

Create a ClusterRole with the aggregate-to-admin label and a nonResourceURLs rule:

apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: test-aggregated-admin-role
  labels:
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
rules:
- nonResourceURLs:
  - '*'
  verbs:
  - get

Apply the ClusterRole:

oc apply -f test-aggregated-admin-role.yaml

Verify that the nonResourceURLs rule is propagated to the aggregated admin ClusterRole:

oc get clusterrole admin -o yaml

Create or reconcile an Argo CD instance managed by the OpenShift GitOps Operator.

Observe that the GitOps Operator attempts to create/update the namespaced Role:

openshift-gitops-argocd-application-controller

The reconciliation fails with:

Role.rbac.authorization.k8s.io "openshift-gitops-argocd-application-controller" is invalid:
rules[814].nonResourceURLs: Invalid value: []string{"*"}:
namespaced rules cannot apply to non-resource URLs

Expected Result

The GitOps Operator should not copy the nonResourceURLs rule into the namespaced Application Controller Role. Argo CD reconciliation should complete successfully.

### DIFF
--- a/test/openshift/e2e/ginkgo/sequential/1-113_validate_controller_role_test.go
+++ b/test/openshift/e2e/ginkgo/sequential/1-113_validate_controller_role_test.go
@@ -56,6 +56,21 @@ func roleContainsPolicyRule(k8sClient client.Client, role *rbacv1.Role, expected
 	return false
 }
 
+func roleContainsNonResourceURLRules(k8sClient client.Client, role *rbacv1.Role) bool {
+	if err := k8sClient.Get(context.Background(), client.ObjectKeyFromObject(role), role); err != nil {
+		GinkgoWriter.Println(err)
+		return true
+	}
+
+	for _, rule := range role.Rules {
+		if len(rule.NonResourceURLs) > 0 {
+			GinkgoWriter.Println("roleContainsNonResourceURLRules - found nonResourceURLs rule:", rule)
+			return true
+		}
+	}
+	return false
+}
+
 var _ = Describe("GitOps Operator Sequential E2E Tests", func() {
 
 	Context("1-113_validate_controller_role", func() {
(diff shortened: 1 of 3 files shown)


akhilnittala (commented):
LGTM

nmirasch (approved):
LGTM!

test/openshift/e2e/ginkgo/sequential/1-113_validate_controller_role_test.go
```diff
@@ -56,6 +56,21 @@ func roleContainsPolicyRule(k8sClient client.Client, role *rbacv1.Role, expected
 	return false
 }
 
+func roleContainsNonResourceURLRules(k8sClient client.Client, role *rbacv1.Role) bool {
+	if err := k8sClient.Get(context.Background(), client.ObjectKeyFromObject(role), role); err != nil {
+		GinkgoWriter.Println(err)
+		return true
```
anandrkskd, line 62: why are we returning true in case of failure, this will make the condition pass even when the k8sClient fails to get the obj.

alkakumari016: We are expecting false from this function for the happy path as the name suggests roleContainsNonResourceURLRules. It will return true either if there is an error or nonResourceUrl is present(which is not expected)

anandrkskd: Understood. make sense. And we are logging the error as well so it will be clear in the logs that the failure happened due to api call hicups and not due to NonResourceURLs found.

anandrkskd (approved):
/lgtm
/approve

PR #1305, opened 2026-09-21
fix(dex): fetching changes for mounting dex sa token as projected volume mount

anandf:
**What type of PR is this?**
/kind bug

Dex server pod restarts the dex server process whenever the static clientSecret changes. This causes existing token verification to fail, as the signing tokens stored in memory gets deleted. To overcome this limitation, external storage options (kubernetes/etcd) was evaluated, though it solved the problem of loosing the dex data, it requires additional care and encryption mechanism to store the sensitive data securely.

Using projected tokens, helps in token rotation directly in the disk by the kubelet and by making an enhancement in dex to read these short-lived tokens directly from the file, helps to avoid the dex server process restarts. It also simplifies the current token rotation mechanism which requires the operator to continuosly look at the service account secrets for any expiry and use the TokenRequest API to generate the token stored in a kubernetes Secret object.

N/A

Fixes
https://redhat.atlassian.net/browse/GITOPS-10364
https://redhat.atlassian.net/browse/GITOPS-10429

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test


akhilnittala (approved):
LGTM

svghadi (approved):
/lgtm
/approve

svghadi (approved):
/lgtm

PR #1307, opened 2026-09-22
fix(redis): Fallback to old AUTH var for the purposes of smooth migration

olivergondza:
After https://github.com/argoproj-labs/argocd-operator/pull/2113, the password has moved from var to file. During HA upgrade, the pods gets the same scripts, but different method of password injection crashing the pre-upgrade pods. The fallback is active only during the migration.

https://redhat.atlassian.net/browse/GITOPS-11294

**What type of PR is this?**

/kind bug
> /kind cleanup
> /kind failing-test
> /kind enhancement
> /kind documentation
> /kind code-refactoring

Fixes #?

**Test acceptance criteria**:

* [ ] Unit Test
* [ ] E2E Test


argocd-operator/build/redis/haproxy_init.sh.tpl
```diff
@@ -33,12 +33,17 @@ if [ -z "$ANNOUNCE_IP2" ]; then
 fi
 sed -i "s/REPLACE_ANNOUNCE2/$ANNOUNCE_IP2/" "$HAPROXY_CONF"
 
-AUTH="$(cat /app/config/redis-auth/auth)"
-if [ -z "${AUTH}" ]; then
+redis_pwd="$(cat /app/config/redis-auth/auth)"
+if [ -z "${redis_pwd}" ]; then
     echo "Error: Redis password not mounted correctly"
-    exit 1
+    if [ ! -z "$AUTH" ]; then
```
anandf, line 39: Where is `AUTH` variable initialized ? Is it expected to be set as environment variable ?

olivergondza: It WAS the old way of passing it before we changed it to the mounted file. The missing file can either indicate a misconfiguration, or that this is one of the pods not yet rolled out during upgrade so the AUTH var is still part of the pod spec. See the JIRA issue for the exact chain of events I believe lead to this regression.

olivergondza:
@akhilnittala, please take a look when you have a minute...

